PR #13, opened 2020-02-05
Transaction id should be optional for correlation info

tomkerkhove:
Transaction id should be optional for correlation info.

In a Web API context we'll generate one, but messaging just uses what's there so this should be optional.


src/Arcus.Observability.Correlation/CorrelationInfo.cs
```diff
@@ -15,9 +14,6 @@ public CorrelationInfo(string operationId, string transactionId)
         {
             Guard.NotNullOrEmpty(operationId, nameof(operationId),
                 "Cannot create a correlation instance with a blank operation ID");
-            Guard.For<ArgumentException>(
```
stijnmoreels: Ok, optional; but do we allow empty? Shouldn't we allow either filled or `null`?

tomkerkhove: Frankly, we should just accept whatever there is - `null`, ``, ` ` or `abc`.

We don't care, do we?

stijnmoreels: It's discussable, and a bit political 😄 , if it's optional or if it's allowed to be empty. You see. 'empty' can be both seen as input or as something that's between non-input and input.
Hence, 'optional' or 'empty'.

stijnmoreels: I guess it depends on the type and the validation of the model.

tomkerkhove: Let me rephrase - We shouldn't care what the value is, it's up to the user to decide if it's valid or not ;)

stijnmoreels: Ah, yes, ok. Shame though that the validation is not done at creation-level. But, yeah. I see your point.

stijnmoreels: So if it's not `null`, then we have a transaction ID?

tomkerkhove: It's something the consumer has to define if it's valid or not so we can't provide.

Frankly, I don't care :D We just pass it, rest is up to them.

tomkerkhove: It's a functional thing, not a technical one

Issue #19, opened 2020-02-06
Provide a Kubernetes enricher


tomkerkhove:
Provide a Kubernetes enricher that automatically adds information about:
- Namespace
- Name of node
- Name of pod

This will be based on environment variables that are added to deployments, similar to the following:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: demo-app-deployment
  labels:
          [... REDACTED ...]
spec:
  replicas: 2
  selector:
    matchLabels:
        app.kubernetes.io/name: demo-app
        app.kubernetes.io/instance: instance
  template:
    metadata:
      labels:
        app.kubernetes.io/name: demo-app
        app.kubernetes.io/instance: instance
    spec:
      containers:
        - name: api
          [... REDACTED ...]
          env:
          - name: KUBERNETES_NODE_NAME
            valueFrom:
              fieldRef:
                fieldPath: spec.nodeName
          - name: KUBERNETES_POD_NAME
            valueFrom:
              fieldRef:
                fieldPath: metadata.name
          - name: KUBERNETES_NAMESPACE
            valueFrom:
              fieldRef:
                fieldPath: metadata.namespace
```

A sample by Tom will be provided as a starting point.

tomkerkhove:
This is open for docs & YAML sample, right?

stijnmoreels:
> This is open for docs & YAML sample, right?

Yes, I think so. Was something you'ld pick up, yes?

tomkerkhove:
Yes, will do it now

PR #30, opened 2020-02-25
FEAT: add application enricher with component name and instance name option

### LINKED ISSUE #21: Provide application component / role enricher
Provide application component / role enricher that will automatically add information concerning the component and the instance name.

- Component name should be defined when configuring the enricher.
- Instance name will either be machine name (when available via https://github.com/serilog/serilog-enrichers-environment) or pod name (when available via https://github.com/arcus-azure/arcus.observability/issues/19)

stijnmoreels:
Provide a fixed component name and a dependent instance name on an existing instance name log property in a new application enricher.

Closes #21 
Relates to #19


src/Arcus.Observability.Telemetry.Serilog/ApplicationEnricher.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        /// <summary>
+        /// Enrich the log event.
+        /// </summary>
+        /// <param name="logEvent">The log event to enrich.</param>
+        /// <param name="propertyFactory">Factory for creating new properties to add to the event.</param>
+        public void Enrich(LogEvent logEvent, ILogEventPropertyFactory propertyFactory)
+        {
+            EnrichComponentName(ComponentName, _componentValue, logEvent, propertyFactory);
+            EnrichInstanceName(logEvent);
```
tomkerkhove: We don't need instance name as this is enriched by other process, see issue:
> Instance name will either be machine name (when available via serilog/serilog-enrichers-environment) or pod name (when available via #19)

src/Arcus.Observability.Tests.Unit/Telemetry/LogEventExtensions.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System.Linq;
+using GuardNet;
+
+// ReSharper disable once CheckNamespace
+namespace Serilog.Events
```
tomkerkhove: Move this in testing or our core?

src/Arcus.Observability.Telemetry.Serilog/ApplicationInstance.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace Arcus.Observability.Telemetry.Serilog 
+{
+    /// <summary>
+    /// Represents which application instance the <see cref="ApplicationEnricher"/> should enrich to the log event properties.
+    /// </summary>
+    public enum ApplicationInstance
```
tomkerkhove: Is this still used?

stijnmoreels: 🤦‍♂

src/Arcus.Observability.Tests.Unit/Telemetry/LogEventExtensions.cs
stijnmoreels: Ah, can you think of scenario's where this would come in handy maybe? Then yeah, sure!

tomkerkhove: Yes, in enrichers and so

tomkerkhove: But ok to keep for now if you prefer

stijnmoreels: I'll move it.

Issue #17, opened 2020-02-06
Provide extension on Serilog to log metrics


tomkerkhove:
Provide extension on Serilog to log metrics:

Ideas are:
```csharp
var context = new Dictionary<string, object>
{
    {ContextProperties.Customer, "Contoso"}
};

_logger.LogMetric("Orders Processed", 1, context);
```

Extension should allow you to provide contextual information by passing a dictionary, but should not be mandatory.

It should seemlessly integrate with the Azure Application Insight sink and track it as a "Metric".

A sample by Tom will be provided as a starting point.

stijnmoreels:
Done?

Issue #16, opened 2020-02-06
Provide extension on Serilog to log dependencies


tomkerkhove:
Provide extension on Serilog to log dependencies:

Ideas are:
```csharp
_logger.LogHttpDependency(requestMessage, statusCode, duration);
_logger.LogSqlDependency(DatabaseInfo.ServerName, DatabaseInfo.DatabaseName, TableName, operationName, isSuccessful, stopwatch.Elapsed);
```

Extension should allow you to provide contextual information by passing a dictionary, but should not be mandatory.

It should seemlessly integrate with the Azure Application Insight sink and track it as a "Dependency" which is nicely rendered in the application map.

A sample by Tom will be provided as a starting point.

stijnmoreels:
Done?

Issue #15, opened 2020-02-06
Provide extension on Serilog to log custom events


tomkerkhove:
Provide extension on Serilog to log custom events:

Ideas are:
```csharp
var context = new Dictionary<string, object>
{
    {ContextProperties.Networking.OldNetworkStatus, oldStatus ?? NetworkStatus.Unknown},
    {ContextProperties.Networking.NetworkStatus, newStatus}
};

_logger.LogEvent("Network Connectivity Changed", context);
```

Extension should allow you to provide contextual information by passing a dictionary, but should not be mandatory.

It should seemlessly integrate with the Azure Application Insight sink and track it as a "Custom Event".

A sample by Tom will be provided as a starting point.

tomkerkhove:
This was done in #40 and documentation is up on https://observability.arcus-azure.net/features/writing-different-telemetry-types#events, ok if we close @stijnmoreels?

stijnmoreels:
> This was done in #40 and documentation is up on https://observability.arcus-azure.net/features/writing-different-telemetry-types#events, ok if we close @stijnmoreels?

Same, should we have some integration tests with Application Insights?

tomkerkhove:
We need to add tests indeed, but only unit tests given we rely on Serilog to handle the rest

stijnmoreels:
Done?

Issue #14, opened 2020-02-06
Provide extension on Serilog to log requests


tomkerkhove:
Provide extension on Serilog to log requests:

Ideas are:
```csharp
logger.LogRequest(httpContext.Request, httpContext.Response, duration);
```

Extension should allow you to provide contextual information by passing a dictionary, but should not be mandatory.

It should seemlessly integrate with the Azure Application Insight sink and track it as a "Request".

A sample by Tom will be provided as a starting point.

stijnmoreels:
Done?

PR #35, opened 2020-03-03
Provide ILogger extensions for emitting request, dependency, metric & event telemetry.

tomkerkhove:
Provide ILogger extensions for emitting request, dependency, metric & event telemetry.

Tests will be provided in a follow-up PR.

Relates to #15 
Relates to #16 
Relates to #17 
Relates to #14


tomkerkhove:
I need your testing advice @stijnmoreels - I need to create an ILogger and get access to the content that was written.

Do we have tests that does this already? If not, is there a way other than writing our own Logger to achieve this?

stijnmoreels:
> I need your testing advice @stijnmoreels - I need to create an ILogger and get access to the content that was written.
> 
> Do we have tests that does this already? If not, is there a way other than writing our own Logger to achieve this?

Yes, we already have tests for this. It uses a `InMemorySink` that concurrently receives the written events which you can get from your test.
For example the version enricher: https://github.com/arcus-azure/arcus.observability/blob/master/src/Arcus.Observability.Tests.Unit/Telemetry/VersionEnricherTests.cs

tomkerkhove:
That's using Serilog which I don't want to use

stijnmoreels:
> That's using Serilog which I don't want to use

Ah, OK, sorry.
You could use a `Mock.Of<>` implementation and "spy" on the written stuff.

tomkerkhove:
It mocks the operation but can I access *what* was written? Anyhow I'll have a look.

stijnmoreels:
> It mocks the operation but can I access _what_ was written? Anyhow I'll have a look.

Something like this I had in mind: https://github.com/Moq/moq4/wiki/Quickstart#callbacks
But again, it's maybe for Arcus users too much.
A dedicated logger that collects the events concurrently could be more useful than this.

src/Arcus.Observability.Telemetry.Core/Arcus.Observability.Telemetry.Core.csproj
```diff
@@ -0,0 +1,28 @@
...
+    <PackageTags>Azure;Observability;Telemetry;Serilog</PackageTags>
+    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <AssemblyName>Arcus.Observability.Telemetry.Core</AssemblyName>
+    <RootNamespace>Arcus.Observability.Telemetry.Core</RootNamespace>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Guard.Net" Version="1.2.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Http.Abstractions" Version="2.2.0" />
```
stijnmoreels: Should we here again use the guidance to use both .NET Core 3.0 and netstandard?

src/Arcus.Observability.Telemetry.Core/ContextProperties.cs
```diff
@@ -0,0 +1,46 @@
+﻿#pragma warning disable 1591
+namespace Arcus.Observability.Telemetry.Core
+{
+    public class ContextProperties
```
stijnmoreels: All static classes? Maybe no need to use XML docs, idk, I'll leave it up to you.

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
```diff
@@ -0,0 +1,133 @@
...
+            context = context ?? new Dictionary<string, object>();
+
+            var requestUri = request.RequestUri;
+            var targetName = requestUri.Host;
+            var requestMethod = request.Method;
+            var dependencyName = $"{requestMethod} {requestUri.AbsolutePath}";
+            var isSuccessful = (int)statusCode >= 200 && (int)statusCode < 300;
+
+            logger.LogInformation(
+                $"{MessagePrefixes.DependencyViaHttp} {{{ContextProperties.DependencyTracking.TargetName}}} for {{{ContextProperties.DependencyTracking.DependencyName}}} completed with {{{ContextProperties.DependencyTracking.ResultCode}}} in {{{ContextProperties.DependencyTracking.Duration}}} at {{{ContextProperties.DependencyTracking.StartTime}}} (Successful: {{{ContextProperties.DependencyTracking.IsSuccessful}}} - Context: {{@{ContextProperties.EventTracking.EventContext}}})",
```
stijnmoreels: The message prefix also in the formatting template instead of concatenating?

src/Arcus.Observability.Telemetry.Core/MessagePrefixes.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace Arcus.Observability.Telemetry.Core
+{
+    public class MessagePrefixes
```
stijnmoreels: Static class?

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
```diff
@@ -0,0 +1,133 @@
...
+using System.Net;
+using System.Net.Http;
+using Arcus.Observability.Telemetry.Core;
+using GuardNet;
+using Microsoft.AspNetCore.Http;
+
+// ReSharper disable once CheckNamespace
+namespace Microsoft.Extensions.Logging
+{
+    public static class ILoggerExtensions
```
stijnmoreels: I guess it's not necessary to also check the key/values in the context dictionary for null?

src/Arcus.Observability.Telemetry.Core/Arcus.Observability.Telemetry.Core.csproj
tomkerkhove: Good point!

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
tomkerkhove: No because it's optional

src/Arcus.Observability.Telemetry.Core/ContextProperties.cs
tomkerkhove: It's for a more fluent approach of using them but open for suggestions

src/Arcus.Observability.Telemetry.Core/MessagePrefixes.cs
tomkerkhove: It's for a more fluent approach of using them but open for suggestions

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
tomkerkhove: This is intentional as we want to use the value here while with the other approach we want to use structured logging.

This results in : `Dependency {DependencyName}` while with your suggestion it would be ``{DependencyType} {DependencyName}`.

if it's ok for you I'd like to keep it as is and revise when I add the converters, is that ok or do you want to change this perse?

stijnmoreels: Aha is good

src/Arcus.Observability.Telemetry.Core/ContextProperties.cs
stijnmoreels: Can't you create static classes inside static classes?

src/Arcus.Observability.Tests.Unit/Telemetry/ILoggerExtensionsTests.cs
```diff
@@ -0,0 +1,274 @@
...
+            // Arrange
+            var logger = new TestLogger();
+            string eventName = null;
+
+            // Act & Arrange
+            Assert.Throws<ArgumentException>(() => logger.LogEvent(eventName));
+        }
+
+        [Fact]
+        public void LogSqlDependency_ValidArguments_Succeeds()
```
stijnmoreels: Nicely tested, easy read and correct assertion. Cool! 🥇

tomkerkhove: I learned from the best!

Issue #23, opened 2020-02-11
Feature - Security event enrichment with either tag and/or property


stijnmoreels:
**What feature would you like to have?**
We enrich the log events with the machine name, context, location, client, ... It would also be beneficial to add a security tag and/or log property. It can be just to tag the event but can also be a level of 'critical/warning/informational'.

When any security event happens (authentication, authorization, input validation, malicious input, additional stuff like firewall/whitelisting, ...) Both the successful and failure could be written with that extra security tag/property.

To oversight the application, it would be very good that these events can be routed. And filtered out. So any malicious activity can be more easily detected.

_Originally posted by @stijnmoreels in https://github.com/arcus-azure/arcus.observability/issues/21#issuecomment-583719302_

stijnmoreels:
An good example is authentication. When the auth for the client fails AND succeeds, and extra security tag/property should be added to the log event so we can route it later.

It can be as simple as:
```csharp
logEvent.AddPropertyIfAbsent("Tag", "Security");
```

But more fine-grained levels, possible with levels of more or less critical can be useful.

tomkerkhove:
Hm, how would you write an enricher that knows when to add the tag and when not?

Isn't it more up to the person logging it to include the tagging as he knows when it's appropriate or not?

stijnmoreels:
Yes, user decides, but we could provide the system. We could even provide a different logging class that adds this (`SecLog.Critical()` for example).
But that's all implementation details.

tomkerkhove:
I'd argue that this is something that is up to the developer to add:
> log.LogInformation("Hello auth ({Tag})", "Security")

We could provide another logger but it should be an extension then.

That said, I don't see how an enricher can help here as it would have to have intelligence to know when to add the property.

stijnmoreels:
> That said, I don't see how an enricher can help here as it would have to have intelligence to know when to add the property.

Yes, when we have a separate logger, a tag should/could be used. But again, that's all implementation details. The point is that with either/both tag/enricher/... or something else... we could easily distinguish security events and that the developer can easily choose whether he/she/x wants to log a security event or not.

tomkerkhove:
That's ok for me, but I think having extensions would be good enough then, shouldn't become too complex.

stijnmoreels:
[...]

PR #40, opened 2020-03-10
Provide Application Insights converters, telemetry type filter and configuration extension

tomkerkhove:
- [X] Provide Application Insights converters
- [X] Provide telemetry type filter
- [X] Provide configuration extension to wire-up the Application Insights sink
- [X] Provide docs for filter
- [X] Provide docs for sink

As discussed with @stijnmoreels, the idea is to have the basics in and add variety of tests in a follow-up PR.

Relates to #14
Relates to #15
Relates to #16
Relates to #17
Relates to #22

PR #42, opened 2020-03-16
Feature - add security logging to 'ILogger' extensions

stijnmoreels:
Relates to  https://github.com/arcus-azure/arcus.observability/issues/23


stijnmoreels:
Not yet docs... first checking if it's OK this way.

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
```diff
@@ -11,6 +13,16 @@ namespace Microsoft.Extensions.Logging
 {
     public static class ILoggerExtensions
     {
+        private const string RequestFormat =
```
stijnmoreels: I think lots of perf improvement here when we only have to create once this format.

tomkerkhove: Definately makes sense, thanks!

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
```diff
@@ -31,12 +43,19 @@ public static void LogRequest(this ILogger logger, HttpRequest request, HttpResp
...
             var host = $"{request.Scheme}://{request.Host}";
 
-            logger.LogInformation(
-                $"{MessagePrefixes.RequestViaHttp} {{{ContextProperties.RequestTracking.RequestMethod}}} {{{ContextProperties.RequestTracking.RequestHost}}}/{{{ContextProperties.RequestTracking.RequestUri}}} completed with {{{ContextProperties.RequestTracking.ResponseStatusCode}}} in {{{ContextProperties.RequestTracking.RequestDuration}}} at {{{ContextProperties.RequestTracking.RequestTime}}} (Context: {{@{ContextProperties.EventTracking.EventContext}}})",
-                request.Method,
-                host, resourcePath, statusCode, duration, DateTimeOffset.UtcNow, context);
+            logger.LogInformation(RequestFormat, request.Method, host, resourcePath, statusCode, duration, DateTimeOffset.UtcNow, context);
         }
 
+        private const string HttpDependencyFormat =
```
tomkerkhove: Can we move all of these to the top please, easier to see all constants that are available

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
```diff
@@ -53,17 +72,25 @@ public static void LogHttpDependency(this ILogger logger, HttpRequestMessage req
 
             context = context ?? new Dictionary<string, object>();
 
-            var requestUri = request.RequestUri;
-            var targetName = requestUri.Host;
-            var requestMethod = request.Method;
-            var dependencyName = $"{requestMethod} {requestUri.AbsolutePath}";
-            var isSuccessful = (int)statusCode >= 200 && (int)statusCode < 300;
+            Uri requestUri = request.RequestUri;
```
tomkerkhove: Good point, thanks! :+1:

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
```diff
@@ -125,9 +159,42 @@ public static void LogMetric(this ILogger logger, string name, double value, Dic
...
+        }
+
+        /// <summary>
+        ///     Logs an event related to an security activity (i.e. input validation, authentication, authorization...).
+        /// </summary>
+        /// <param name="logger">The logger to use.</param>
+        /// <param name="categoryName">The name of the security category.</param>
+        /// <param name="message">The user message written with in the <paramref name="categoryName"/>.</param>
+        /// <param name="arguments">The arguments formatted with the <paramref name="message"/>.</param>
+        public static void LogSecurityEvent(this ILogger logger, string categoryName, string message, params object[] arguments)
```
tomkerkhove: In my opinion, we should guide our consumers to write good traces so we should enforce this via our parameters.

Now we are just mapping it directly on what .NET enforces us to give

tomkerkhove: Think what we need is:
- Security event name
- Details about the event

Something like this? But it's hard to come up with this without a concrete case

stijnmoreels: For example something I think about during input validation:

```csharp
logger.LogError("{SecurityCategory} - input is invalid person's age, > {MaxAge}", "Input validation", 130);
```

stijnmoreels: That's also one of the reasons I initially said we could use an enricher, bc I can think of different type of scenario's with different kind of critical levels, with different kind of arguments to send,...

tomkerkhove: Not sure why that's a security event imo but I see your point.

In that case we have a security category, message and parameters, thinkthat should cut it. You cannot use an enricher for this kind of stuff because enrichers have no knowledge of the scenario, they just enrich based on existing messages 🤷‍♂

tomkerkhove:
If you add some docs we can merge this

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
stijnmoreels: It's about diagnostics you know, if you see for example an input validation mutiple times, see a pattern... you can figure out more easily malicious activity. And with adding an extra security property, makes the filtering more easily.

stijnmoreels:
> If you add some docs we can merge this

Waiting for https://github.com/arcus-azure/arcus.observability/pull/40 so we have extend the telemetry doc.

stijnmoreels:
> If you add some docs we can merge this

Should I remove the arguments in the signature and use a `Context` instead? Making use of the already `LogEvent`?

tomkerkhove:
That would be good indeed but if you use `LogEvent` then it should not have the prefix but more add it to the context with `EventType: Security` instead

tomkerkhove:
Would allow you to provide consistent and clean approach and yet be able to filter out all security events

src/Arcus.Observability.Tests.Unit/Telemetry/ILoggerExtensionsTests.cs
```diff
@@ -277,33 +277,16 @@ public void LogSecurityEvent_ValidArguments_Succeeds()
         {
```
tomkerkhove: Can we add a test which uses the context as well please?

Issue #18, opened 2020-02-06
Move to correlation enricher to arcus.observability


tomkerkhove:
Move to correlation enricher which already lives in [arcus.webapi](https://github.com/arcus-azure/arcus.webapi/pull/99) to arcus.observability

stijnmoreels:
The one in the web API repo uses the HTTP correlation. That's probably not what we want?

tomkerkhove:
That's what we've discussed before - We'll find a way to make it work for both messaging & web API or provide 2 flavors.

tomkerkhove:
Discussed this again and today, ideally, the HTTP correlation information (or at least the generic info) is available via the dependency container so we can build it upon that.

tomkerkhove:
You could also argue that this should be part of every seperate project since messaging, for example, has a cycle id which needs to be added as well.

Maybe that makes more sense then  @stijnmoreels?

stijnmoreels:
Some more information on where and how this correlation should be available. Parallel requrests with different correlation, places where correlation is needed...

tomkerkhove:
### Background

Let's take as step back and look at what we want to achieve first.

We'd like to provide a consistent approach on how we perform correlation in our applications where we are mainly thinking in terms of Web APIs & workers (messaging).

As part of that effort; we have identified and carved out [`CorrelationInfo`](https://github.com/arcus-azure/arcus.observability/blob/master/src/Arcus.Observability.Correlation/CorrelationInfo.cs) which contains `TransactionId` & `OperationId` which are used by both.

Both Web APIs & workers (messaging) have these concepts while messaging takes it a step further given it also has the concept of `CycleId`.

Our goal is to provide one or more enrichers which will automatically extend all of our telemetry with all of the correlation information that is available

### Challenge

Our challenge is to provide a simple and clean way to achieve this. However, it's not simple because:
- Web API provides [`CorrelationMiddleware`](https://github.com/arcus-azure/arcus.webapi/blob/master/src/Arcus.WebApi.Correlation/CorrelationMiddleware.cs) which registers `CorrelationInfo` as an HTTP feature
- Messaging reads the correlation information inside the message pump and passes it to the message handlers

Both do not register the information in the dependency container so we cannot access it like this.

### Options

1. We move the correlation to dependency container as scoped so that we can write one enricher based on that info.
2. We provide multiple enrichers and make them specific for the application type that need them. No need to add them to the dependency container, but more stuff to add

[...]

PR #43, opened 2020-03-16
Feature - move correlation access information to observability

stijnmoreels:
Relates to https://github.com/arcus-azure/arcus.observability/issues/18


stijnmoreels:
Hang on for the docs until this is approved.

src/Arcus.Observability.Correlation/IServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// <summary>
+        /// Adds operation and transaction correlation to the application.
+        /// </summary>
+        /// <param name="services">The services collection containing the dependency injection services.</param>
+        /// <param name="createCustomCorrelationAccessor">The custom <see cref="ICorrelationInfoAccessor"/> implementation factory to retrieve the <see cref="CorrelationInfo"/>.</param>
+        /// <param name="configureOptions">The function to configure additional options how the correlation works.</param>
+        public static IServiceCollection AddCorrelation<TAccessor>(
+            this IServiceCollection services, 
+            Func<IServiceProvider, TAccessor> createCustomCorrelationAccessor,
+            Action<CorrelationInfoOptions> configureOptions = null)
```
stijnmoreels: Should we also provide a generic type for the options so specific application correlation configurations can be configured here as well?

stijnmoreels:
When this is merged, we can in the Web Api repo use the `ICorrelationInfoAccessor` to implement it instead of using now the `HttpCorrelationInfo`.
In short, the `HttpCorrelationInfo` will implement this interface and we could use the interface across the application.

src/Arcus.Observability.Correlation/IServiceCollectionExtensions.cs
stijnmoreels: Or... is the options here moved also too application specific.

stijnmoreels: I'm a bit of expirimenting with maybe even make the `CorrelationInfo` type generic, so we can use directly (for example) the `MessageCorrelationInfo` instead of casting; when in a messaging context.

src/Arcus.Observability.Correlation/ICorrelationInfoAccessorT.cs
```diff
@@ -0,0 +1,14 @@
...
+    /// <summary>
+    /// Represents a contract to access the <typeparamref name="TCorrelationInfo"/> model.
+    /// </summary>
+    public interface ICorrelationInfoAccessor<TCorrelationInfo> 
+        where TCorrelationInfo : CorrelationInfo
+    {
+        /// <summary>
+        /// Gets or sets the current correlation information initialized in this context.
+        /// </summary>
+        TCorrelationInfo CorrelationInfo { get; set; }
```
tomkerkhove: We should only support get maybe or do you want to use this to automatically set all the required bits, for example in our correlationd middleware?

Maybe less relevant?

src/Arcus.Observability.Telemetry.Serilog.Enrichers/CorrelationInfoEnricher.cs
```diff
@@ -0,0 +1,56 @@
...
+            {
+                LogEventProperty property = propertyFactory.CreateProperty(ContextProperties.Correlation.OperationId, correlationInfo.OperationId);
+                logEvent.AddPropertyIfAbsent(property);
+            }
+
+            if (!String.IsNullOrEmpty(correlationInfo.TransactionId))
+            {
+                LogEventProperty property = propertyFactory.CreateProperty(ContextProperties.Correlation.TransactionId, correlationInfo.TransactionId);
+                logEvent.AddPropertyIfAbsent(property);
+            }
```
tomkerkhove: This was what I was wondering actually - How would you approach it to get messaging correlation info then? Wouldn't it make more sense to provide an enricher per app type then?

tomkerkhove: For the rest, 👌

src/Arcus.Observability.Tests.Unit/Correlation/TestCorrelationInfoAccessor.cs
```diff
@@ -0,0 +1,20 @@
...
+    /// Test <see cref="ICorrelationInfoAccessor{TCorrelationInfo}"/> implementation to access the <see cref="TestCorrelationInfo"/> model
+    /// </summary>
+    public class TestCorrelationInfoAccessor : ICorrelationInfoAccessor<TestCorrelationInfo>
+    {
+        /// <summary>
+        /// Gets or sets the current correlation information initialized in this context.
+        /// </summary>
+        public TestCorrelationInfo CorrelationInfo
+        {
+            get => throw new NotImplementedException();
```
tomkerkhove: 🤔

src/Arcus.Observability.Correlation/ICorrelationInfoAccessor.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Arcus.Observability.Correlation 
+{
+    /// <summary>
+    /// Represents a contract to access the <see cref="CorrelationInfo"/> model.
+    /// </summary>
+    public interface ICorrelationInfoAccessor : ICorrelationInfoAccessor<CorrelationInfo>
```
tomkerkhove: Why do we have this interface? Just for convenience

src/Arcus.Observability.Correlation/CorrelationInfoAccessorProxy.cs
```diff
@@ -0,0 +1,35 @@
+﻿using GuardNet;
+
+namespace Arcus.Observability.Correlation
+{
+    internal class CorrelationInfoAccessorProxy<TCorrelationInfo> : ICorrelationInfoAccessor
```
tomkerkhove: Why do we need this?

src/Arcus.Observability.Correlation/DefaultCorrelationInfoAccessor.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System.Threading;
+
+namespace Arcus.Observability.Correlation
+{
+    /// <summary>
+    /// Default <see cref="ICorrelationInfoAccessor"/> implementation to access the <see cref="Correlation.CorrelationInfo"/> in the current context.
+    /// </summary>
+    public class DefaultCorrelationInfoAccessor : ICorrelationInfoAccessor
+    {
+        private static readonly AsyncLocal<CorrelationInfo> CorrelationInfoLocalData = new AsyncLocal<CorrelationInfo>();
```
tomkerkhove: Is this the idea to get rid of the current web api approach and use this instead or?

src/Arcus.Observability.Correlation/CorrelationInfoAccessorProxy.cs
stijnmoreels: To register a `ICorrelationInfoAccessor<>` as a `ICorrelationInfoAccessor`.

src/Arcus.Observability.Correlation/DefaultCorrelationInfoAccessor.cs
stijnmoreels: This is the default accessor for I think the messaging repo.
The web app can implement it's own accessor and use the HTTP request features instead.

src/Arcus.Observability.Correlation/ICorrelationInfoAccessor.cs
stijnmoreels: Yes convenience and for the applications that dont need a special correla info model.

src/Arcus.Observability.Correlation/ICorrelationInfoAccessorT.cs
stijnmoreels: I thought about it, but the messaging repo is gonna need some setting of the correlation on some way.
This get/set seems like a solution.

src/Arcus.Observability.Tests.Unit/Correlation/TestCorrelationInfoAccessor.cs
stijnmoreels: Just for testing. In the right terminology it would be called a "dummy".http://xunitpatterns.com/Dummy%20Object.html

stijnmoreels: Do you want me to change it to a regular property?

src/Arcus.Observability.Correlation/DefaultCorrelationInfoAccessor.cs
stijnmoreels: Just as it uses now, only now instead of the `HttpCorrelationInfo` we have there, we'll have a `HttpCorrelationInfoAccessor`.

src/Arcus.Observability.Correlation/CorrelationInfoAccessorProxy.cs
stijnmoreels: It's really about convenience.

src/Arcus.Observability.Telemetry.Serilog.Enrichers/CorrelationInfoEnricher.cs
stijnmoreels: Hmm, maybe we can indeed make some inject point in the enricher where the user can send for a custom correlation accessor a custom function. 🤔
I'll come up with something.

src/Arcus.Observability.Tests.Unit/Correlation/TestCorrelationInfoAccessor.cs
tomkerkhove: Since you're not using it I would use a regular property instead indeed

src/Arcus.Observability.Correlation/DefaultCorrelationInfoAccessor.cs
tomkerkhove: Sounds good to me! Only concern I have is that correlation info for messaging is defined on a pre-message basis so does it work with this appraoch?

stijnmoreels: I was thinking that we could maybe set the correlation info at the same place where we call `message.GetMessageCorrelationInfo()` in the message pump.
We could retrieve the `ICorrelationInfo<MessageCorrelationInfo>` from the `.ServiceProvider` property in the message pump and set the correlation.

src/Arcus.Observability.Telemetry.Serilog.Enrichers/CorrelationInfoEnricher.cs
stijnmoreels: Updated with function now... is it something you could agree upon?

src/Arcus.Observability.Correlation/DefaultCorrelationInfoAccessor.cs
tomkerkhove: That's ok for me, but I would do it after calling GetMessageCorrelationInfo which is probably what you are suggesting as well?

src/Arcus.Observability.Telemetry.Serilog.Enrichers/CorrelationInfoEnricher.cs
tomkerkhove: Instead of a function I would make it a method which can be overriden; this is more the approach used in .NET

tomkerkhove: But fixes the issue indeed, thanks!

src/Arcus.Observability.Telemetry.Serilog.Enrichers/Extensions/LoggerEnrichmentConfigurationExtensions.cs
```diff
@@ -44,5 +48,52 @@ public static LoggerConfiguration WithKubernetesInfo(this LoggerEnrichmentConfig
...
+        }
+
+        /// <summary>
+        /// Adds the <see cref="CorrelationInfoEnricher{TCorrelationInfo}"/> to the logger enrichment configuration which adds the custom <see cref="CorrelationInfo"/> information from the current context.
+        /// </summary>
+        /// <typeparam name="TCorrelationInfo">The type of the custom <see cref="CorrelationInfo"/> model.</typeparam>
+        /// <param name="enrichmentConfiguration">The configuration to add the enricher.</param>
+        /// <param name="correlationInfoAccessor">The accessor implementation for the <see cref="CorrelationInfo"/> model.</param>
+        /// <param name="enrichAdditionalCorrelationInfo">The function to enrich additional information from the custom <see cref="CorrelationInfo"/> model</param>
+        public static LoggerConfiguration WithCorrelationInfo<TCorrelationInfo>(
```
tomkerkhove: I'd not use  this enrich action, has to be done by providing a different enricher instead of the default one

src/Arcus.Observability.Correlation/DefaultCorrelationInfoAccessor.cs
stijnmoreels: Yes, indeed. After. Ok, we're on the same page.

src/Arcus.Observability.Telemetry.Serilog.Enrichers/CorrelationInfoEnricher.cs
stijnmoreels: 😆 I guess it's the functional programmer in me to imm think of functions.

stijnmoreels:
I removed the extra extension to add the corelation enricher.
Since we create a new custom enricher for custom correlation info models, it seems a bit useless to create a dedicated extension that does only `Enrich.With`.

src/Arcus.Observability.Telemetry.Serilog.Enrichers/Extensions/LoggerEnrichmentConfigurationExtensions.cs
```diff
@@ -44,5 +48,34 @@ public static LoggerConfiguration WithKubernetesInfo(this LoggerEnrichmentConfig
 
             return enrichmentConfiguration.With<KubernetesEnricher>();
         }
+
```
tomkerkhove: With the model of extending CorrelationInfoEnricher by deriving from it we should give them the capability to pass it around here or use that one instead (I wish there was a `: new(<>)` :()

docs/features/telemetry-enrichment.md
```diff
@@ -73,4 +73,43 @@ ILogger logger = new LoggerConfiguration()
...
+Name: `TransactionId`
+Value: `0477E377-414D-47CD-8756-BCBE3DBE3ACB`
+
+**Usage**
+
+```csharp
+ICorrelationInfoAccessor defaultImplementation = new DefaultCorrelationInfoAccessor();
+
+ILogger logger = new LoggerConfiguration()
+    .Enrich.WithCorrelationInfo(defaultImplementation)
```
tomkerkhove: Seeing this, I think we should provide an overload `.Enrich.WithCorrelationInfo()` that uses `DefaultCorrelationInfoAccessor` behind the scenes

docs/features/telemetry-enrichment.md
```diff
@@ -73,4 +73,43 @@ ILogger logger = new LoggerConfiguration()
...
+    .CreateLogger();
+
+logger.Information("This event will be enriched with the correlation information");
```
tomkerkhove: You might want to show the output of this

docs/features/correlation.md
```diff
@@ -18,4 +18,124 @@ This feature requires to install our NuGet package
...
+        // (default: new `Guid`).
+        options.Operation.GenerateId = () => $"Operation-{Guid.NewGuid()}";
+    });
+}
+```
+
+Later in the application, the options can be retrieved by injecting the `IOptions<CorrelationInfoOptions>` type.
+
+[&larr; back](/)
+=======
```
tomkerkhove:
```suggestion

```

docs/features/correlation.md
```diff
@@ -18,4 +18,124 @@ This feature requires to install our NuGet package
...
 [&larr; back](/)
+>>>>>>> master
```
tomkerkhove:
```suggestion

```

docs/features/correlation.md
```diff
@@ -18,4 +18,124 @@ This feature requires to install our NuGet package
...
+        options.Operation.GenerateId = () => $"Operation-{Guid.NewGuid()}";
+    });
+}
+```
+
+Later in the application, the options can be retrieved by injecting the `IOptions<CorrelationInfoOptions>` type.
+
+[&larr; back](/)
+=======
 [&larr; back](/)
```
tomkerkhove:
```suggestion

```

docs/features/correlation.md
```diff
@@ -18,4 +18,124 @@ This feature requires to install our NuGet package
 PM > Install-Package Arcus.Observability.Correlation
 ```
 
+## What We Provide
+
+The `Arcus.Observability.Correlation` library provides a way to get access to correlation information across your application.
+What it **DOES NOT** provide is how this correlation information is initially set.
```
tomkerkhove: We changed this, no?

tomkerkhove:
Looks pretty good to me, only thing I'd suggest would be to create a PR on Web API using the package of this PR to show what it would look  like over there. Bonus points if we do this for Messaging as well.

Or implement them here, but via a dedicated PR to this one; maybe that's even better.

stijnmoreels:
> Looks pretty good to me, only thing I'd suggest would be to create a PR on Web API using the package of this PR to show what it would look like over there. Bonus points if we do this for Messaging as well.
> 
> Or implement them here, but via a dedicated PR to this one; maybe that's even better.

Was already doing that a bit.

docs/features/telemetry-enrichment.md
stijnmoreels: Ah, yes, thank you.

stijnmoreels: Hmm, the problem exists that we don't have access anymore to that default accessor. Unless you want me to make it a public available singleton (`DefaultCorrelationInfoAccessor.Instance`).

src/Arcus.Observability.Telemetry.Serilog.Enrichers/Extensions/LoggerEnrichmentConfigurationExtensions.cs
stijnmoreels: It's not very much added value maybe (only `Enrich.With()` i's going to contain).

docs/features/telemetry-enrichment.md
stijnmoreels: Done just that.

docs/features/correlation.md
stijnmoreels: Well, the place where you set the correlation is specific to the application, no? (Web Api = middleware + HTTP features, Message pump = OnReceive)  So, I would leave it.

stijnmoreels:
> > Looks pretty good to me, only thing I'd suggest would be to create a PR on Web API using the package of this PR to show what it would look like over there. Bonus points if we do this for Messaging as well.
> > Or implement them here, but via a dedicated PR to this one; maybe that's even better.
> 
> Was already doing that a bit.

>> https://github.com/arcus-azure/arcus.webapi/pull/132

src/Arcus.Observability.Correlation/IServiceCollectionExtensions.cs
stijnmoreels: Added generic `TOptions` for this.

docs/features/telemetry-enrichment.md
```diff
@@ -110,6 +110,8 @@ ILogger logger = new LoggerConfiguration()
     .CreateLogger();
 
 logger.Information("This event will be enriched with the correlation information");
+// Output: This event will be enriched with the correlation information
```
tomkerkhove: I'd output it as it is in really so most probably this should be a oneliner?

stijnmoreels:
With what we have now in this PR, is it enough to merge and look how we would introduce it further in the web api and messaging?

src/Arcus.Observability.Tests.Unit/Serilog/TestCorrelationInfoEnricher.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// </summary>
+        /// <param name="logEvent">The log event to enrich with correlation information.</param>
+        /// <param name="propertyFactory">The log property factory to create log properties with correlation information.</param>
+        /// <param name="correlationInfo">The correlation model that contains the current correlation information.</param>
+        protected override void EnrichCorrelationInfo(
+            LogEvent logEvent,
+            ILogEventPropertyFactory propertyFactory,
+            TestCorrelationInfo correlationInfo)
+        {
+            LogEventProperty property = propertyFactory.CreateProperty(TestId, correlationInfo.TestId);
```
tomkerkhove: Are you deliberately ignore the base here?

docs/features/telemetry-enrichment.md
tomkerkhove: What have you done? Provide an overload where you don't have to pass it or?

If so, please update the docs here.

tomkerkhove:
I've added a few comments and merging is ok but let's wait a while before releasing 0.1 so that we can make further changes if need be

src/Arcus.Observability.Tests.Unit/Serilog/TestCorrelationInfoEnricher.cs
stijnmoreels: Yes, I guess we could test both...

tomkerkhove: Fine by me, just wanted to verify!

stijnmoreels:
Updated the docs with the default impl like you said. Is indeed without an argument now.

Issue #71, opened 2020-04-01
Verify custom dependency & request integration tests


tomkerkhove:
Verify custom dependency & request integration tests.

Follow-up of https://github.com/arcus-azure/arcus.observability/pull/62 & https://github.com/arcus-azure/arcus.observability/pull/64

stijnmoreels:
Why we can't see the custom dependencies is still something I don't understand. Is there something wrong with the names of the log properties? Bc there's no exception, no wrong message template, no spaces in dependency type... And I'm sure I should see it on the application map, right?

stijnmoreels:
Maybe some required properties here 🤔 but then the fault is hidden somewhere. I'll look into that...

tomkerkhove:
> And I'm sure I should see it on the application map, right?

You should also see them if you go to `Search` and filter on `Event Type: Dependency`:
![image](https://user-images.githubusercontent.com/4345663/78770564-063a7500-798f-11ea-9d7c-493694009c92.png)

If you go to the App Map, make sure to increase the time period in which you'd expect them to show up:
![image](https://user-images.githubusercontent.com/4345663/78770639-21a58000-798f-11ea-86a0-f8d11718fabb.png)

Here you see why it's important to use the correct dependency type as they use it to visualize the dependency with (ie SQL & HTTP)

You are probably asking because of the custom dependency tests then or?

tomkerkhove:
You can apply filters as well in search:
![image](https://user-images.githubusercontent.com/4345663/78770903-92e53300-798f-11ea-84c1-3d5a67c62db7.png)

stijnmoreels:
Yes, asking about the custom dependency tests. Nothing to show in Application Insights 😭 .

stijnmoreels:
Made sure every property was filled out so I don't miss required properties.

PR #72, opened 2020-04-01
Fix - tracking Application Insights telemetry

stijnmoreels:
Relates to #71

### DIFF
--- a/src/Arcus.Observability.Tests.Integration/Serilog/ApplicationInsightsSinkTests.cs
+++ b/src/Arcus.Observability.Tests.Integration/Serilog/ApplicationInsightsSinkTests.cs
@@ -103,7 +103,7 @@ namespace Arcus.Observability.Tests.Integration.Serilog
                 ILogger logger = loggerFactory.CreateLogger<ApplicationInsightsSinkTests>();
 
                 HttpMethod httpMethod = GenerateHttpMethod();
-                string host = _bogusGenerator.Company.CompanyName();
+                string host = _bogusGenerator.Company.CompanyName().Replace(" ", "");
                 string path = "/" + _bogusGenerator.Commerce.ProductName();
                 HttpRequest request = CreateStubRequest(httpMethod, "https", host, path);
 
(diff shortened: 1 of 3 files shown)


src/Arcus.Observability.Tests.Integration/Serilog/ApplicationInsightsSinkTests.cs
```diff
@@ -103,7 +103,7 @@ public void LogRequest_SinksToApplicationInsights_ResultsInRequestTelemetry()
                 ILogger logger = loggerFactory.CreateLogger<ApplicationInsightsSinkTests>();
 
                 HttpMethod httpMethod = GenerateHttpMethod();
-                string host = _bogusGenerator.Company.CompanyName();
+                string host = _bogusGenerator.Company.CompanyName().Replace(" ", "");
```
tomkerkhove, line 106: Why is this required? Does it cause any issues?

stijnmoreels: Yes, spaces in host names are now allowed.

tomkerkhove: Let's enforce it in our extension then

stijnmoreels: The extension already uses `HttpRequest` instead so that's already covered, right? The tests only uses this for the stubbing.

tomkerkhove: Does HttpRequest throw an exception for that host name or not? If not, we should add a check in our extension to enforce it maybe?

stijnmoreels: I don't think it guards it very well, so I'll add some extra guards here.

tomkerkhove: Ok thanks!

Issue #58, opened 2020-03-30
Provide capability to track an Azure Service Bus dependency


tomkerkhove:
**Is your feature request related to a problem? Please describe.**
Provide capability to track an Azure Service Bus dependency.

**Describe the solution you'd like**
We should provide a `LogServiceBusDependency` overload for `ILogger` which is optimized for Azure Service Bus and correctly generates the [Azure Application Insights App Map](https://docs.microsoft.com/en-gb/azure/azure-monitor/app/app-map).

It should support both queues and topics.

**Describe alternatives you've considered**
None

**Additional context**
Dependency type should be used from [official SDK](https://github.com/microsoft/ApplicationInsights-dotnet/blob/develop/WEB/Src/DependencyCollector/Shared/Implementation/RemoteDependencyConstants.cs).

stijnmoreels:
On what available properties do you think this overload should have?
- Entity type queue/topic
- Entity Name
- ...?

tomkerkhove:
Let's get it working first and then see what we need :)

stijnmoreels:
Tried something to also investigate in #71 but nothing so far 😭 .

stijnmoreels:
Done?

PR #76, opened 2020-04-09
Feature - add Azure Service Bus tracking

stijnmoreels:
Relates to #58


src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
```diff
@@ -31,6 +31,14 @@ public static class ILoggerExtensions
             + ContextProperties.DependencyTracking.IsSuccessful + "} - Context: {@"
             + ContextProperties.EventTracking.EventContext + "})";
 
+        private const string ServiceBusDependencyFormat =
+            MessagePrefixes.Dependency + " Azure Service Bus {"
```
tomkerkhove, line 35: Do we use this approach, or do we create a converter like others?

tomkerkhove: See https://github.com/arcus-azure/arcus.observability/pull/76/files?file-filters%5B%5D=.cs#diff-75d51c6e532953a0a10af8789970abeeR25

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
```diff
@@ -109,6 +135,36 @@ public static void LogDependency(this ILogger logger, string dependencyType, obj
...
 
+        /// <summary>
+        ///     Logs an Azure Service Bus Dependency.
+        /// </summary>
+        /// <param name="logger">Logger to use</param>
+        /// <param name="dependencyType">Type of the Service Bus entity</param>
+        /// <param name="isSuccessful">Indication whether or not the operation was successful</param>
+        /// <param name="measurement">Measuring the latency to call the Service Bus dependency</param>
+        /// <param name="context">Context that provides more insights on the dependency that was measured</param>
+        public static void LogServiceBusDependency(this ILogger logger, ServiceBusDependencyType dependencyType, bool isSuccessful, DependencyMeasurement measurement, Dictionary<string, object> context = null)
```
tomkerkhove, line 146: Thinking out loud:
- Do we need to make the Service Bus dependency type optional? For example, do we always know if we are sending to a queue or topic?
- Instead of making it a parameter, what if it's just called `LogServiceBusQueueDependency`?

I would go for having a dedicated method per type (`LogServiceBusQueueDependency` & `LogServiceBusTopicDependency`) and provide `LogServiceBusDependency` which doesn't support it maybe?

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
tomkerkhove: I think having it would be ideal because I'd like to add `EntityType` as a context property which will be `Queue`, `Topic` or `Unknown` (when not specified, see below)

stijnmoreels: 1. I like the way we now reuse existing custom dependency functionality, actually. So the distinction is made at client level but not internally. Especially if there would be no extra functionality.
2. Yes, that would be good! I was wondering actually which dependency tracking properties we should use for this, what their values should be, ...

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
stijnmoreels: Yes, I can do that (the last proposal).

stijnmoreels: Done that.

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
stijnmoreels: 2. Done.

docs/features/writing-different-telemetry-types.md
```diff
@@ -30,10 +30,47 @@ Dependencies allow you to track how your external dependencies are doing to give
...
+    { "Namespace", "azure.servicebus.namespace" }
+};
+
+var durationMeasurement = new Stopwatch();
+
+// Start measuring
+durationMeasurement.Start();
+var startTime = DateTimeOffset.UtcNow;
+
+_logger.LogServiceBusDependency(ServiceBusDependencyType.Queue, isSuccessful: true, startTime, durationMeasurement.Elapsed, telemetryContext);
```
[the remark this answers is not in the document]
stijnmoreels: Updated the log to use the `TargetName` for that and updated the docs.

src/Arcus.Observability.Tests.Integration/Serilog/ApplicationInsightsSinkTests.cs
```diff
@@ -143,6 +143,27 @@ public void LogDependency_SinksToApplicationInsights_ResultsInDependencyTelemetr
...
+            {
+                ILogger logger = loggerFactory.CreateLogger<ApplicationInsightsSinkTests>();
+
+                string entityName = _bogusGenerator.Commerce.Product();
+                bool isSuccessful = _bogusGenerator.PickRandom(true, false);
+                TimeSpan duration = _bogusGenerator.Date.Timespan();
+                Dictionary<string, object> telemetryContext = CreateTestTelemetryContext();
+
+                // Act
+                logger.LogServiceBusDependency(entityName, isSuccessful, DateTimeOffset.UtcNow, duration, ServiceBusEntityType.Queue, telemetryContext);
```
tomkerkhove, line 160: This is not showing up in Application Insights; at least not correctly

tomkerkhove, line 160: Can we align the timestamp with other tests?

src/Arcus.Observability.Tests.Unit/Serilog/ApplicationInsightsTelemetryConverterTests.cs
```diff
@@ -122,6 +122,53 @@ public void LogDependency_WithDependency_CreatesDependencyTelemetry()
...
+            AssertDoesNotContainLogProperty(logEvent, DependencyTracking.DependencyName);
+            AssertDoesNotContainLogProperty(logEvent, DependencyTracking.IsSuccessful);
+            AssertDoesNotContainLogProperty(logEvent, DependencyTracking.DependencyData);
+            AssertDoesNotContainLogProperty(logEvent, DependencyTracking.ResultCode);
+            AssertDoesNotContainLogProperty(logEvent, DependencyTracking.StartTime);
+            AssertDoesNotContainLogProperty(logEvent, DependencyTracking.Duration);
+            Assert.Collection(telemetries, telemetry =>
+            {
+                var dependencyTelemetry = Assert.IsType<DependencyTelemetry>(telemetry);
+                Assert.Equal("Azure Resource", dependencyTelemetry.Type);
```
tomkerkhove, line 161: This should be `Azure Service Bus`
```suggestion
                Assert.Equal("Azure Service Bus", dependencyTelemetry.Type);
```

src/Arcus.Observability.Tests.Unit/Serilog/ApplicationInsightsTelemetryConverterTests.cs
```diff
@@ -122,6 +122,53 @@ public void LogDependency_WithDependency_CreatesDependencyTelemetry()
             });
         }
 
+        [Fact]
+        public void LogServiceBusDependency_WithServiceBusDependency_CreatesDependencyTelemetry()
```
tomkerkhove, line 126: Can we add tests for our methods who are specific for queue & topic please? One overload is ok

src/Arcus.Observability.Tests.Unit/Telemetry/ILoggerExtensionsTests.cs
```diff
@@ -99,6 +99,172 @@ public void LogDependency_ValidArguments_Succeeds()
...
+            measurement.Dispose();
+            TimeSpan duration = measurement.Elapsed;
+
+            // Act
+            logger.LogServiceBusDependency(entityName, isSuccessful, measurement, dependencyType);
+
+            // Assert
+            string logMessage = logger.WrittenMessage;
+            Assert.StartsWith(MessagePrefixes.Dependency, logMessage);
+            Assert.Contains(dependencyType.ToString(), logMessage);
```
tomkerkhove, line 170: Dependency type should be `Azure Service Bus`, not the entity type.

src/Arcus.Observability.Tests.Unit/Telemetry/ILoggerExtensionsTests.cs
```diff
@@ -99,6 +99,172 @@ public void LogDependency_ValidArguments_Succeeds()
...
+            DateTimeOffset startTime = measurement.StartTime;
+            measurement.Dispose();
+            TimeSpan duration = measurement.Elapsed;
+
+            // Act
+            logger.LogServiceBusTopicDependency(topicName, isSuccessful, measurement);
+
+            // Assert
+            string logMessage = logger.WrittenMessage;
+            Assert.StartsWith(MessagePrefixes.Dependency, logMessage);
```
tomkerkhove, line 261: We should extend these asserts in all tests to ensure that it's followed by `Azure Service Bus`

src/Arcus.Observability.Tests.Unit/Telemetry/ILoggerExtensionsTests.cs
```diff
@@ -99,6 +99,172 @@ public void LogDependency_ValidArguments_Succeeds()
...
+            bool isSuccessful = _bogusGenerator.PickRandom(true, false);
+            TimeSpan duration = _bogusGenerator.Date.Timespan();
+            var startTime = DateTimeOffset.UtcNow;
+
+            // Act
+            logger.LogServiceBusDependency(entityName, isSuccessful, startTime, duration, dependencyType);
+
+            // Assert
+            string logMessage = logger.WrittenMessage;
+            Assert.StartsWith(MessagePrefixes.Dependency, logMessage);
```
tomkerkhove, line 144: Please add assert in all tests to ensure that the `entityName` is included

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
```diff
@@ -31,6 +31,16 @@ public static class ILoggerExtensions
             + ContextProperties.DependencyTracking.IsSuccessful + "} - Context: {@"
             + ContextProperties.EventTracking.EventContext + "})";
 
+        private const string ServiceBusDependencyFormat =
+            MessagePrefixes.Dependency + " Azure Service Bus {"
```
tomkerkhove, line 35: `Azure Service Bus` is the dependency type to use

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
```diff
@@ -31,6 +31,16 @@ public static class ILoggerExtensions
             + ContextProperties.DependencyTracking.IsSuccessful + "} - Context: {@"
             + ContextProperties.EventTracking.EventContext + "})";
 
+        private const string ServiceBusDependencyFormat =
```
tomkerkhove, line 34: Honestly, I'm not sure why we can't re-use `DependencyFormat` if we are not adding a new converter?

stijnmoreels: Now we use the `TargetName` which is not in the dependency format.

tomkerkhove: Can't we add it then to our regular dependency tracking as well? Everybody should be able to specify that.

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
stijnmoreels: Changed that.

src/Arcus.Observability.Tests.Unit/Telemetry/ILoggerExtensionsTests.cs
stijnmoreels: Changed that.

src/Arcus.Observability.Tests.Unit/Telemetry/ILoggerExtensionsTests.cs
stijnmoreels: Done that.

src/Arcus.Observability.Tests.Unit/Telemetry/ILoggerExtensionsTests.cs
stijnmoreels: Done that.

src/Arcus.Observability.Tests.Unit/Serilog/ApplicationInsightsTelemetryConverterTests.cs
stijnmoreels: Done that.

src/Arcus.Observability.Tests.Unit/Serilog/ApplicationInsightsTelemetryConverterTests.cs
stijnmoreels: Yes, changed that.

src/Arcus.Observability.Tests.Integration/Serilog/ApplicationInsightsSinkTests.cs
stijnmoreels: I think it shows up now correctly.

stijnmoreels: Yep, done that.

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
stijnmoreels: We can make an issue of that to add the `TargetName` to that one, but what about the `EntityType` that is now added to the message format?

src/Arcus.Observability.Tests.Integration/Serilog/ApplicationInsightsSinkTests.cs
tomkerkhove: I'm still seeing the old `Azure Resource`, did you see anything else?
image

Also not that there are properties that should be gone

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
tomkerkhove: That's ok, if we get rid of the SB one then or align it

src/Arcus.Observability.Tests.Integration/Serilog/ApplicationInsightsSinkTests.cs
stijnmoreels: image

docs/features/writing-different-telemetry-types.md
```diff
@@ -30,10 +30,47 @@ Dependencies allow you to track how your external dependencies are doing to give
...
+var telemetryContext = new Dictionary<string, object>
+{
+    { "Namespace", "azure.servicebus.namespace" }
+};
+
+// Start measuring
+using (var measurement = DependencyMeasurement.Start())
+{
+    _logger.LogServiceBusQueueDependency(queueName: "ordersqueue", isSuccessful: true, measurement, telemetryContext);
+    // Output: "Dependency Azure ServiceBus Queue named ordersqueue in 00:00:00.2521801 at 03/23/2020 09:56:31 +00:00 (Successful: True - Context: [Namespace, azure.servicebus.namespace])"
```
tomkerkhove, line 70: Shouldn't it be this instead?
```suggestion
    // Output: "Dependency Azure Service Bus Queue named ordersqueue in 00:00:00.2521801 at 03/23/2020 09:56:31 +00:00 (Successful: True - Context: [Namespace, azure.servicebus.namespace])"
```

docs/features/writing-different-telemetry-types.md
```diff
@@ -30,10 +30,47 @@ Dependencies allow you to track how your external dependencies are doing to give
...
+    { "Namespace", "azure.servicebus.namespace" }
+};
+
+var durationMeasurement = new Stopwatch();
+
+// Start measuring
+durationMeasurement.Start();
+var startTime = DateTimeOffset.UtcNow;
+
+_logger.LogServiceBusQueueDependency(queueName: "ordersqueue", isSuccessful: true, startTime, durationMeasurement.Elapsed, telemetryContext);
```
tomkerkhove, line 54: We should highlight that we have an overload for topic or generic as well

docs/features/writing-different-telemetry-types.md
stijnmoreels: Ah, yes, thanks.

docs/features/writing-different-telemetry-types.md
stijnmoreels: Yes, add a note about that.

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
stijnmoreels: What do you mean exactly? 🤔

docs/features/writing-different-telemetry-types.md
tomkerkhove: Did the outcome change as well or was it just a doc issue?

src/Arcus.Observability.Tests.Integration/Serilog/ApplicationInsightsSinkTests.cs
tomkerkhove: :+1: Can you removes these 2 properties as well along wiht `DependencyStartTime` & `DependencyDuration` please?

Also, why does it say "EventDescription"? Because it's a dependency?

stijnmoreels: That's the name we use for the context on all log types.

stijnmoreels: Those properties are by default added to custom dependencies too. Just reusing functionality here.

docs/features/writing-different-telemetry-types.md
stijnmoreels: No, was indeed something I missed in the docs.

docs/features/writing-different-telemetry-types.md
```diff
@@ -30,10 +30,49 @@ Dependencies allow you to track how your external dependencies are doing to give
...
 We provide support for the following dependencies:
 
+- [Azure Service Bus](#measuring-azure-service-bus-dependencies)
 - [Custom](#measuring-custom-dependencies)
 - [HTTP](#measuring-http-dependencies)
 - [SQL](#measuring-sql-dependencies)
 
+### Measuring Azure Service Bus dependencies
+
+Here is how you can report an Azure Service Bus dependency:
```
tomkerkhove, line 40: Maybe it's better here to mention what we support instead of diving into an example immediately?

For example:
```suggestion

We allow you to measure Azure Service Bus dependencies for both queues & topics.

Here is how you can report an Azure Service Bus Queue dependency:
```

docs/features/writing-different-telemetry-types.md
tomkerkhove: Added suggestion above as well, think both are the sweet spot.

src/Arcus.Observability.Tests.Integration/Serilog/ApplicationInsightsSinkTests.cs
tomkerkhove: I think we should get rid of them as the information is already there and more noise making it harder to find what you need

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
tomkerkhove: Nevermind, will wait for that PR

src/Arcus.Observability.Tests.Integration/Serilog/ApplicationInsightsSinkTests.cs
tomkerkhove: Can be via a dedicated PR if you want, but don't forget to create an issue then.

stijnmoreels: In both the custom dependency as in the service bus dependency?

docs/features/writing-different-telemetry-types.md
stijnmoreels: Yes, and also unknown entities.

src/Arcus.Observability.Tests.Integration/Serilog/ApplicationInsightsSinkTests.cs
tomkerkhove: Yes because we just use them to pass it to the converter so that he can set the correct properties unless I'm missing something ofcourse.

stijnmoreels: Addd an issue for this: https://github.com/arcus-azure/arcus.observability/issues/81

docs/features/writing-different-telemetry-types.md
```diff
@@ -30,10 +30,51 @@ Dependencies allow you to track how your external dependencies are doing to give
...
 We provide support for the following dependencies:
 
+- [Azure Service Bus](#measuring-azure-service-bus-dependencies)
 - [Custom](#measuring-custom-dependencies)
 - [HTTP](#measuring-http-dependencies)
 - [SQL](#measuring-sql-dependencies)
 
+### Measuring Azure Service Bus dependencies
+
+We allow you to measure Azure Service Bus dependencies for both queues & topics or unknown entities.
```
tomkerkhove, line 40: I get what you mean but here it doesn't really make sense if you ask me.
```suggestion
We allow you to measure Azure Service Bus dependencies for both queues & topics.
```

You'll always know the entity but maybe you don't know its type; but don't think it's important to emphasize that here. The reason why I wanted to add this line is clearly state that it both supports queues & topics which are part of the same product.

PR #82, opened 2020-04-15
Feature - remove intermediary log properties before forwarding to telemetry

### LINKED ISSUE #81: Remove unnecessary custom properties from the dependency tracking
Some custom properties are set to create a `DependencyTelemetry` instance in the `CustomDependencyTelemetryConverter`, but they remain there when they should actually be removed.

These properties should be removed after the creation of the `DependencyTelemetry` instance:
- `DependencyDuration`
- `DependencyStartTime`

stijnmoreels:
Closes #81

### DIFF
--- a/src/Arcus.Observability.Telemetry.Serilog.Sinks.ApplicationInsights/Converters/CustomTelemetryConverter.cs
+++ b/src/Arcus.Observability.Telemetry.Serilog.Sinks.ApplicationInsights/Converters/CustomTelemetryConverter.cs
@@ -30,9 +30,9 @@ namespace Arcus.Observability.Telemetry.Serilog.Sinks.ApplicationInsights.Conver
             AssignTelemetryContextProperties(logEvent, telemetryEntry);
             _cloudContextConverter.EnrichWithAppInfo(logEvent, telemetryEntry);
 
+            RemoveIntermediaryProperties(logEvent);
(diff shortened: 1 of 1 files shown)


stijnmoreels:
Cleaner details now:
image

stijnmoreels:
> LGTM, one question - What is `EventDescription`? I presume this is passed via context properties?

That's I think the telemetry context I think.

tomkerkhove:
Nitpick - Can we rename it?

stijnmoreels:
> Nitpick - Can we rename it?

Yes, we can, it will effect other telemetry too of course. `EventContext`?

stijnmoreels:
We can remove it too, of course, since it's a duplication.

tomkerkhove:
I thought it was passed here: https://github.com/arcus-azure/arcus.observability/blob/master/src/Arcus.Observability.Tests.Integration/Serilog/ApplicationInsightsSinkTests.cs#L162

But looks like it's coming from https://github.com/arcus-azure/arcus.observability/blob/e8872f93b4d0b30691e14e71a0732caed69d0675/src/Arcus.Observability.Telemetry.Core/ContextProperties.cs#L32

Moving this to https://github.com/arcus-azure/arcus.observability/issues/83 so this can be closed.

PR #85, opened 2020-04-17
Testing - add integration tests for Application Insights sink

### LINKED ISSUE #61: Provide integration tests that write to Azure Application Insights via Serilog
Provide integration tests that write to Azure Application Insights via Serilog.

Provide integration tests that write to Azure Application Insights via Serilog.

Every test should include following context:
- Generate & include correlation information
- Name of test
- Name of test run (if possible at all but not a must)

This would allow us to use it to lookup the written telemetry during verification if we want to.

**Additional context**
This allows us to manually verify the information in Azure Application Insights before shipping, for example is the app map correctly rendered?

stijnmoreels:
Closes #61

### DIFF
--- a/src/Arcus.Observability.Tests.Integration/Serilog/ApplicationInsightsSinkTests.cs
+++ b/src/Arcus.Observability.Tests.Integration/Serilog/ApplicationInsightsSinkTests.cs
@@ -432,17 +432,17 @@ namespace Arcus.Observability.Tests.Integration.Serilog
         }
 
         [Fact]
-        public void LogHttpDependencyWithComponentName_SinksToApplicationInsights_ResultsInHttpDependencyTelemetryWithComponentName()
+        public async Task LogHttpDependencyWithComponentName_SinksToApplicationInsights_ResultsInHttpDependencyTelemetryWithComponentName()
         {
             // Arrange
             string componentName = _bogusGenerator.Commerce.ProductName();
+            HttpMethod httpMethod = GenerateHttpMethod();
+            string requestUrl = _bogusGenerator.Image.LoremFlickrUrl();
             using (ILoggerFactory loggerFactory = CreateLoggerFactory(config => config.Enrich.WithComponentName(componentName)))
             {
                 ILogger logger = loggerFactory.CreateLogger<ApplicationInsightsSinkTests>();
 
-                HttpMethod httpMethod = GenerateHttpMethod();
-                string requestUri = _bogusGenerator.Image.LoremFlickrUrl();
-                var request = new HttpRequestMessage(httpMethod, requestUri)
+                var request = new HttpRequestMessage(httpMethod, requestUrl)
                 {
                     Content = new StringContent(_bogusGenerator.Lorem.Paragraph())
                 };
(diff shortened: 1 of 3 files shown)


stijnmoreels:
@tomkerkhove , can you make the Azure DevOps required in the build on this repo? I don't have enough permissions.

tomkerkhove:
What do you want me to do?

stijnmoreels:
> What do you want me to do?

Making it a requirement in the branch policies.

tomkerkhove:
Make what a requirement? 😅

stijnmoreels:
> Make what a requirement?

Currently the Azure DevOps build of the Observability repo is not a requirement in the branch policy of the `master`,  so would you add this? Or make me admin so I can do it? Now it's not required to succeed to merge.

tomkerkhove:
We don't need to do that since it will enforce it for all PRs, including doc only changes.

The build defines that already

stijnmoreels:
> We don't need to do that since it will enforce it for all PRs, including doc only changes.
> 
> The build defines that already

Ok, then we'll leave it like this. Are the tests enough to be approved?

src/Arcus.Observability.Tests.Integration/Serilog/ApplicationInsightsSinkTests.cs
```diff
@@ -237,43 +322,66 @@ public void LogSqlDependency_SinksToApplicationInsights_ResultsInSqlDependencyTe
...
                 // Act
-                logger.LogEvent("Update version");
-
-                // Assert
-                // Hold on till we have agreed on assertion...
+                logger.LogEvent(eventName);
             }
+
+            // Assert
+            // Hold on till we have agreed on assertion...
```
tomkerkhove, line 384: Is this on purpose that we don't cover it?

src/Arcus.Observability.Tests.Integration/Serilog/ApplicationInsightsSinkTests.cs
```diff
@@ -3,28 +3,38 @@
...
 using Serilog.Extensions.Logging;
 using Xunit;
 using Xunit.Abstractions;
+using Xunit.Sdk;
 using ILogger = Microsoft.Extensions.Logging.ILogger;
 
 namespace Arcus.Observability.Tests.Integration.Serilog
 {
     [Trait(name: "Category", value: "Integration")]
     public class ApplicationInsightsSinkTests : IntegrationTest
```
tomkerkhove, line 33: This looks good to me, I'm just wondering if we should include correlation to ensure that it's the specific entry we want to have, but I'll leave it up to you to judge.

stijnmoreels: By using generated values, and including them in the assert, we're pretty sure that it's the correct one we're after. Several of them also include a time range of 1h to limit this even further. The correletion we could add will probably not be available in the retrieved entries anyway bc they are custom properties.

src/Arcus.Observability.Tests.Integration/Serilog/ApplicationInsightsSinkTests.cs
stijnmoreels: As I previously told you on Teams, the problem I had is that the retrieved entries doesn't contain (yet, it's a preview NuGet package) custom properties, so also no enriched properties. What I'm thinking of what we could do, is altering the message template with a test enricher if the log event got enriched with something we want to test.

src/Arcus.Observability.Tests.Integration/Serilog/ApplicationInsightsSinkTests.cs
tomkerkhove: Sounds good to me, but having dedicated test for the correlation enricher to verify it works might be interesting, no?

Just thinking out loud.

src/Arcus.Observability.Tests.Integration/Serilog/ApplicationInsightsSinkTests.cs
tomkerkhove: I think it's ok how it is in that case, resolving!

src/Arcus.Observability.Tests.Integration/Serilog/ApplicationInsightsSinkTests.cs
stijnmoreels: Moved to #93 where we can talk about how.

PR #97, opened 2020-04-29
Feature - run integration tests in parallel

### LINKED ISSUE #96: Are the test running in parallel?
**What feature would you like to have?**
The current build time is arround 36m 40s , and this is probably bc the tests are not run in parallel.

This should be checked.

_Originally posted by @tomkerkhove in https://github.com/arcus-azure/arcus.observability/pull/94#issuecomment-620496248_

stijnmoreels:
Moved related concepts to separate classes, since xUnit makes test collections of test classes or assemblies and not of separate tests. So splitting them, gives xUnit the opportiunity to run them in parallel.

Closes #96


tomkerkhove:
Makes sense!

Only thing I would suggest is to move them from `src/Arcus.Observability.Tests.Integration/Serilog/ApplicationInsightsSinkCustomDependencyTests.cs` to `src/Arcus.Observability.Tests.Integration/Serilog/Sinks/ApplicationInsights/CustomDependencyTests.cs`

That way it's more structured and smaller names, what do you think?

stijnmoreels:
Total of 9min build time. Lot better I'ld say, still under the _10min build time_ rule 😉.

Issue #59, opened 2020-03-30
Provide capability to track an Azure Event Hubs dependency


tomkerkhove:
**Is your feature request related to a problem? Please describe.**
Provide capability to track an Azure Event Hubs dependency.

**Describe the solution you'd like**
We should provide a `LogEventHubsDependency` overload for `ILogger` which is optimized for Azure Event Hubs Storage and correctly generates the [Azure Application Insights App Map](https://docs.microsoft.com/en-gb/azure/azure-monitor/app/app-map).

**Describe alternatives you've considered**
None

**Additional context**
Dependency type should be used from [official SDK](https://github.com/microsoft/ApplicationInsights-dotnet/blob/develop/WEB/Src/DependencyCollector/DependencyCollector/Implementation/RemoteDependencyConstants.cs).

stijnmoreels:
Reopened for docs.

PR #98, opened 2020-04-29
Feature - add Azure Event Hubs dependency tracking

stijnmoreels:
Relates to #59

### DIFF
--- a/src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
+++ b/src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
@@ -349,39 +349,39 @@ namespace Microsoft.Extensions.Logging
         ///     Logs an Azure Event Hub Dependency.
         /// </summary>
         /// <param name="logger">Logger to use</param>
-        /// <param name="accountName">Account of the resource</param>
+        /// <param name="namespaceName">Namespace of the resource</param>
         /// <param name="eventHubName">Name of the Event Hub resource</param>
         /// <param name="isSuccessful">Indication whether or not the operation was successful</param>
         /// <param name="measurement">Measuring the latency to call the dependency</param>
         /// <param name="context">Context that provides more insights on the dependency that was measured</param>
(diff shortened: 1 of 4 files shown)


src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
```diff
@@ -340,9 +340,50 @@ public static void LogTableStorageDependency(this ILogger logger, string account
...
+        ///     Logs an Azure Event Hub Dependency.
+        /// </summary>
+        /// <param name="logger">Logger to use</param>
+        /// <param name="accountName">Account of the resource</param>
+        /// <param name="eventHubName">Name of the Event Hub resource</param>
+        /// <param name="isSuccessful">Indication whether or not the operation was successful</param>
+        /// <param name="startTime">Point in time when the interaction with the dependency was started</param>
+        /// <param name="duration">Duration of the operation</param>
+        /// <param name="context">Context that provides more insights on the dependency that was measured</param>
+        public static void LogEventHubsDependency(this ILogger logger, string accountName, string eventHubName, bool isSuccessful, DateTimeOffset startTime, TimeSpan duration, Dictionary<string, object> context = null)
```
tomkerkhove, line 376: Event Hubs uses the concept of namespaces similar to Service Bus, so we should name it accordingly.
```suggestion
        public static void LogEventHubsDependency(this ILogger logger, string namespaceName, string eventHubName, bool isSuccessful, DateTimeOffset startTime, TimeSpan duration, Dictionary<string, object> context = null)
```

stijnmoreels: And can I use the `namespaceName` as dependency data?

stijnmoreels: Updated.

stijnmoreels:
image

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
tomkerkhove: Sure!

Issue #56, opened 2020-03-30
Provide capability to track an Azure Blob Storage dependency


tomkerkhove:
**Is your feature request related to a problem? Please describe.**
Provide capability to track an Azure Blob Storage dependency.

**Describe the solution you'd like**
We should provide a `LogBlobStorageDependency` overload for `ILogger` which is optimized for Azure Blob Storage and correctly generates the [Azure Application Insights App Map](https://docs.microsoft.com/en-gb/azure/azure-monitor/app/app-map).

**Describe alternatives you've considered**
None

**Additional context**
Dependency type should be used from [official SDK](https://github.com/microsoft/ApplicationInsights-dotnet/blob/develop/WEB/Src/DependencyCollector/DependencyCollector/Implementation/RemoteDependencyConstants.cs).

PR #101, opened 2020-04-30
Feature - add Azure Blob Container dependency tracking

stijnmoreels:
Relates to #56

### DIFF
--- a/src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
+++ b/src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
@@ -315,6 +315,46 @@ namespace Microsoft.Extensions.Logging
         }
 
         /// <summary>
+        ///     Logs an Azure Blob Storage Dependency.
+        /// </summary>
+        /// <param name="logger">Logger to use</param>
+        /// <param name="accountName">Account of the storage resource</param>
+        /// <param name="blobName">Name of the Blob Container resource</param>
+        /// <param name="isSuccessful">Indication whether or not the operation was successful</param>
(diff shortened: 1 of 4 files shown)


src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
```diff
@@ -314,6 +314,46 @@ public static void LogServiceBusDependency(this ILogger logger, string entityNam
...
+        ///     Logs an Azure Blob Storage Dependency.
+        /// </summary>
+        /// <param name="logger">Logger to use</param>
+        /// <param name="accountName">Account of the storage resource</param>
+        /// <param name="blobName">Name of the Blob Container resource</param>
+        /// <param name="isSuccessful">Indication whether or not the operation was successful</param>
+        /// <param name="startTime">Point in time when the interaction with the dependency was started</param>
+        /// <param name="duration">Duration of the operation</param>
+        /// <param name="context">Context that provides more insights on the dependency that was measured</param>
+        public static void LogBlobStorageDependency(this ILogger logger, string accountName, string blobName, bool isSuccessful, DateTimeOffset startTime, TimeSpan duration, Dictionary<string, object> context = null)
```
tomkerkhove, line 346: Not sure what they expect but I would change this to the container name instead

tomkerkhove:
I don't see it pop up on the App Map or is it just me? 🤔

stijnmoreels:
> I don't see it pop up on the App Map or is it just me? 🤔

image

The start time is randomized till 24h, so you got to change the timespan in which the app map is shown.

tomkerkhove:
Ok thanks!

PR #132, opened 2020-07-07
Provide note how to use an optional instrumentation key

### LINKED ISSUE #125: Not specifying an InstrumentationKey for the Azure Application Insights sink throws an ArgumentException
When I configure SeriLog and want to make use of the ApplicationInsights sink, an ArgumentException is thrown when I pass in an empty instrumentation key.

Although I could understand the reasoning, I wonder if this is the behavior that we want.   When using Application Insights directly, one could also instantiate a TelemetryClient and pass in an empty instrumentationkey without an exception being thrown.
I think we should adhere to that behavior.

I have for instance a use case where I've an containerized API which is deployed in Azure, and the same API is also deployed on a boat.   Since the ship does not always have a connection to the Internet, I'm not logging to AppInsights in that case, so my configuration doesn't contain an instrumentation key.  In that case, an exception is thrown when the Serilog logger is being instantiated.

tomkerkhove:
Closes #125

### DIFF
--- a/docs/preview/features/sinks/azure-application-insights.md
+++ b/docs/preview/features/sinks/azure-application-insights.md
@@ -15,7 +15,7 @@ PM > Install-Package Arcus.Observability.Telemetry.Serilog.Sinks.ApplicationInsi
 
 ## What is it?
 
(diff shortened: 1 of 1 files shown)


tomkerkhove:
Let me know what you think @fgheysels

stijnmoreels:
I think it's a nice middle ground here. Cool!

PR #151, opened 2020-11-04
Provide capability to only specifiy status code for LogRequest instead of full response

tomkerkhove:
Provide capability to only specifiy status code for LogRequest instead of full response.


stijnmoreels (approved):
Looks good! Maybe also test the guards, especially bc there's some logic there for whitespace.

stijnmoreels (approved):
Great! Super.

PR #160, opened 2020-11-19
Fix - set correlation info throughout async operations

### LINKED ISSUE #159: Changes to correlation information in `DefaultCorrelationInfoAccessor` are not returned
When using the default correlation info accessor, we set given correlation information multiple times but when retrieving it back; it always provides the initial information, not the last defined on.

Retrieve the latest defined correlation information.

**Additional context**
We are using Azure Functions and Arcus Observability v1.0 by opting in as following:
```csharp
public override void Configure(IFunctionsHostBuilder builder)
{
   builder.Services.AddHttpClient();
   builder.Services.AddCorrelation();
}
```

However, I doubt that this is unrelated to Azure Functions.

Simple workaround for now is:
```csharp
public class InMemoryCorrelationInfoAccessor : ICorrelationInfoAccessor
{
   private CorrelationInfo _correlationInfo;
   public CorrelationInfo GetCorrelationInfo()
   {
       return _correlationInfo;
   }

   public void SetCorrelationInfo(CorrelationInfo correlationInfo)
   {
       _correlationInfo = correlationInfo;
   }
}

public override void Configure(IFunctionsHostBuilder builder)
{
   builder.Services.AddHttpClient();
   builder.Services.AddCorrelation(provider => new InMemoryCorrelationInfoAccessor());
}
```

stijnmoreels:
We first uses a different correlation information for each asyhchronous operation, but this is actually wrong as we already use a scoped service for this. We should consider not using the static value for the default correlation info accessor as this would be the same accross requests. So we should create a new accessor for each request instead.

Closes #159

### DIFF
--- a/src/Arcus.Observability.Telemetry.Serilog.Enrichers/Extensions/LoggerEnrichmentConfigurationExtensions.cs
+++ b/src/Arcus.Observability.Telemetry.Serilog.Enrichers/Extensions/LoggerEnrichmentConfigurationExtensions.cs
@@ -136,6 +136,30 @@ namespace Serilog
         }
 
         /// <summary>
+        /// Adds the previously registered <see cref="ICorrelationInfoAccessor"/> to the logger enrichment configuration which adds the <see cref="CorrelationInfo"/> information from the current context.
+        /// </summary>
+        /// <param name="enrichmentConfiguration">The configuration to add the enricher.</param>
+        /// <param name="serviceProvider">The instance to provide the <see cref="ICorrelationInfoAccessor"/> service while enriching the log events with the correlation information.</param>
+        /// <param name="operationIdPropertyName">The name of the property to enrich the log event with the correlation operation ID.</param>
+        /// <param name="transactionIdPropertyName">The name of the property to enrich the log event with the correlation transaction ID.</param>
+        /// <exception cref="ArgumentNullException">Thrown when the <paramref name="enrichmentConfiguration"/> or <paramref name="serviceProvider"/> is <c>null</c>.</exception>
+        /// <exception cref="ArgumentException">Thrown when the <paramref name="operationIdPropertyName"/> or <paramref name="transactionIdPropertyName"/> is blank.</exception>
+        public static LoggerConfiguration WithCorrelationInfo(
+            this LoggerEnrichmentConfiguration enrichmentConfiguration,
+            IServiceProvider serviceProvider,
+            string operationIdPropertyName = ContextProperties.Correlation.OperationId,
+            string transactionIdPropertyName = ContextProperties.Correlation.TransactionId)
+        {
+            Guard.NotNull(enrichmentConfiguration, nameof(enrichmentConfiguration), "Requires an enrichment configuration to add the correlation information enricher");
+            Guard.NotNull(serviceProvider, nameof(serviceProvider), "Requires a provider to retrieve the correlation information accessor instance");
+            Guard.NotNullOrWhitespace(operationIdPropertyName, nameof(operationIdPropertyName), "Requires a property name to enrich the log event with the correlation operation ID");
+            Guard.NotNullOrWhitespace(transactionIdPropertyName, nameof(transactionIdPropertyName), "Requires a property name to enrich the log event with the correlation transaction ID");
+
+            var accessor = serviceProvider.GetRequiredService<ICorrelationInfoAccessor>();
+            return WithCorrelationInfo(enrichmentConfiguration, accessor, operationIdPropertyName, transactionIdPropertyName);
+        }
+
+        /// <summary>
         /// Adds the <see cref="DefaultCorrelationInfoAccessor{TCorrelationInfo}"/> to the logger enrichment configuration which adds the <see cref="CorrelationInfo"/> information from the current context.
         /// </summary>
         /// <param name="enrichmentConfiguration">The configuration to add the enricher.</param>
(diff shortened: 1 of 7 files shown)


src/Arcus.Observability.Telemetry.Serilog.Enrichers/Extensions/LoggerEnrichmentConfigurationExtensions.cs
```diff
@@ -142,6 +143,7 @@ public static LoggerConfiguration WithVersion(this LoggerEnrichmentConfiguration
         /// <param name="transactionIdPropertyName">The name of the property to enrich the log event with the correlation transaction ID.</param>
         /// <exception cref="ArgumentNullException">Thrown when the <paramref name="enrichmentConfiguration"/> is <c>null</c>.</exception>
         /// <exception cref="ArgumentException">Thrown when the <paramref name="operationIdPropertyName"/> or <paramref name="transactionIdPropertyName"/> is blank.</exception>
+        [Obsolete("Use the " + nameof(WithCorrelationInfo) + " overload with providing your own default correlation accessor")]
```
tomkerkhove, line 146: Hm why are we deprecating this? I don't want to have to provide a correlation info accessor every time myself, no?

src/Arcus.Observability.Tests.Unit/Correlation/DefaultCorrelationInfoAccessorTests.cs
```diff
@@ -0,0 +1,36 @@
...
+            // Act
+            await SetCorrelationInfo(secondOperationId, transactionId);
+
+            // Assert
+            CorrelationInfo correlationInfo = DefaultCorrelationInfoAccessor.Instance.GetCorrelationInfo();
+            Assert.Equal(secondOperationId, correlationInfo.OperationId);
+            Assert.Equal(transactionId, correlationInfo.TransactionId);
+        }
+
+        private async Task SetCorrelationInfo(string operationId, string transactionId)
```
tomkerkhove, line 30: Maybe use a void since nothing here is async?
```suggestion
        private void SetCorrelationInfo(string operationId, string transactionId)
```

tomkerkhove (commented):
Added a few suggestions

src/Arcus.Observability.Tests.Unit/Correlation/DefaultCorrelationInfoAccessorTests.cs
stijnmoreels: No, this was explicitly done bc the `AsyncLocal` was using a different value for each async operation. This made sure I could reproduce this.

src/Arcus.Observability.Telemetry.Serilog.Enrichers/Extensions/LoggerEnrichmentConfigurationExtensions.cs
stijnmoreels: Well, it's static, so we will have the same correlation info accesor across multiple requests (possible with a different correlation information values). Which is not correct. We could return a new accessor on each get-call; but then we wouldn't be able to use that same accessor instance in the dependency container.

tomkerkhove: As a consumer I should not have to care about all of this but just say that you want ot use correlation. So if that means we need to resolve the accessor through the DI container, than we should provide that as a parameter imo.

stijnmoreels: Yes, you're right. I'll add an extension for this.

tomkerkhove:
Forgot to hit send 😅

src/Arcus.Observability.Telemetry.Serilog.Enrichers/Extensions/LoggerEnrichmentConfigurationExtensions.cs
stijnmoreels: You mean also deprecate the extension where you pass along the `ICorrelationInfoAccessor` yourself and guide them to the one with the `IServiceProvider`? Is there not some use cases where it would make sense to pass along the accessor yourself?

tomkerkhove: Not really because with the approach where it has to be passed people would get the impression they have to do something while with the IServiceProvider we manage everything for them?

src/Arcus.Observability.Telemetry.Serilog.Enrichers/Extensions/LoggerEnrichmentConfigurationExtensions.cs
```diff
@@ -134,6 +135,30 @@ public static LoggerConfiguration WithVersion(this LoggerEnrichmentConfiguration
...
+            IServiceProvider serviceProvider,
+            string operationIdPropertyName = ContextProperties.Correlation.OperationId,
+            string transactionIdPropertyName = ContextProperties.Correlation.TransactionId)
+        {
+            Guard.NotNull(enrichmentConfiguration, nameof(enrichmentConfiguration), "Requires an enrichment configuration to add the correlation information enricher");
+            Guard.NotNull(serviceProvider, nameof(serviceProvider), "Requires a provider to retrieve the correlation information accessor instance");
+            Guard.NotNullOrWhitespace(operationIdPropertyName, nameof(operationIdPropertyName), "Requires a property name to enrich the log event with the correlation operation ID");
+            Guard.NotNullOrWhitespace(transactionIdPropertyName, nameof(transactionIdPropertyName), "Requires a property name to enrich the log event with the correlation transaction ID");
+
+            var accessor = serviceProvider.GetRequiredService<ICorrelationInfoAccessor>();
```
tomkerkhove, line 158: Just to verify, this will continue to work fine because people used AddCorrelation before this, right?

tomkerkhove (approved):

src/Arcus.Observability.Telemetry.Serilog.Enrichers/Extensions/LoggerEnrichmentConfigurationExtensions.cs
stijnmoreels: Yes, that's the idea. ☺️

PR #163, opened 2020-12-04
Feature - track Azure Key Vault dependency

### LINKED ISSUE #157: Provide capability to track an Azure Key Vault dependency
Provide the capability to track an Azure Key Vault dependency

stijnmoreels:
Adds a new telemetry extension to track Azure Key Vault dependencies in Application Insights.
This one can be used, later, within the tracking of the secret store (which now uses the general dependency tracking).

Closes #157

### DIFF
--- a/docs/preview/features/writing-different-telemetry-types.md
+++ b/docs/preview/features/writing-different-telemetry-types.md
@@ -38,6 +38,7 @@ We provide support for the following dependencies:
 - [Azure Cosmos DB](#measuring-azure-cosmos-db-dependencies)
 - [Azure Event Hubs](#measuring-azure-event-hubs-dependencies)
 - [Azure IoT Hub](#measuring-azure-iot-hub-dependencies)
+- [Azure Key Vault](#measuring-azure-key-vault-dependencies)
 - [Azure Search](#measuring-azure-search-dependencies)
 - [Azure Service Bus](#measuring-azure-service-bus-dependencies)
 - [Azure Table Storage](#measuring-azure-table-storage-dependencies)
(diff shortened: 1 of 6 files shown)


docs/preview/features/writing-different-telemetry-types.md
```diff
@@ -140,10 +141,29 @@ var durationMeasurement = new Stopwatch();
...
+Here is how you can report a dependency call:
+
+```csharp
+var durationMeasurement = new StopWatch();
+
+// Start measuring
+durationMeasurement.Start();
+var startTime = DateTimeOffset.UtcNow;
+
+logger.AzureKeyVaultDependency(vaultUri: "https://my-secret-store.vault.azure.net", operationName: "get secret", isSuccessful: true, startTime: startTime, duration: durationMeasurement.Elapsed);
```
tomkerkhove, line 163: This should also include the secret name that was consumed

stijnmoreels: Besides the 'operation name' or instead of the operation name ?

stijnmoreels: I removed the operation name and replaced it with the secret name as it will probably make more sense to pass along the name of the secret then then name of the operation (like 'getting secrets' or 'setting secret').

tomkerkhove, line 163: I think that's fine for now indeed; let's just update it here:
```suggestion
logger.AzureKeyVaultDependency(vaultUri: "https://my-secret-store.vault.azure.net", secretName: "ServiceBus-ConnectionString", isSuccessful: true, startTime: startTime, duration: durationMeasurement.Elapsed);
```

tomkerkhove (approved):

PR #166, opened 2020-12-18
Fix - move ASP.NET Core request tracking to seperate library

### LINKED ISSUE #162: Remove ASP.NET dependency in `Arcus.Observability.Telemetry.Core`
When you have a containerized daemon application (non ASP.NET related), and want to use the Arcus ApplicationInsights sink for serilog, you're required to use an asp.net based container image, since the dependencies in `Arcus.Observability.Telemetry.Serilog.Sinks.ApplicationInsights` require the ASP.NET framework.

There exists an ApplicationInsights package that has no dependency on ASP.NET which can be used in these situations: https://docs.microsoft.com/en-us/azure/azure-monitor/app/worker-service

It would be nice if Arcus could provide a sink that builds upon this package and can be used in these situations.

stijnmoreels:
Extracts the `.LogRequest` logger extensions with the ASP.NET Core `HttpRequest`/`HttpResponse` overload to a separate library so the `.Core` library isn't polluted with additional HTTP dependencies.

Closes #162


stijnmoreels:
@fgheysels did I get the target frameworks right this time?

src/Arcus.Observability.Telemetry.AspNetCore/Extensions/ILoggerExtensions.cs
```diff
@@ -0,0 +1,67 @@
...
+using Arcus.Observability.Telemetry.Core;
+using GuardNet;
+using Microsoft.AspNetCore.Http;
+
+// ReSharper disable once CheckNamespace
+namespace Microsoft.Extensions.Logging
+{
+    public static class ILoggerExtensions
+    {
+        private const string RequestFormat =
```
tomkerkhove, line 13: What if we move this to a constant that everyone can use in case they want?

That way you can also re-use it here

stijnmoreels: Could be useful, yes. Mostly useful for us I think but yeah, I'll change that. Thanks.

src/Arcus.Observability.Telemetry.AspNetCore/Arcus.Observability.Telemetry.AspNetCore.csproj
```diff
@@ -0,0 +1,38 @@
...
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <AssemblyName>Arcus.Observability.Telemetry.AspNetCore</AssemblyName>
+    <RootNamespace>Arcus.Observability.Telemetry.AspNetCore</RootNamespace>
+  </PropertyGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.1'">
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' != 'netcoreapp3.1'">
```
fgheysels, line 26: Wouldn't it be better to explicitly check on netstandard2 here ?
Because, what if we're using it in .NET 5 ?  Don't know what will happen then ?  I assume that .NET 5 also depends on Microsoft.AspnetCore.App, but I haven't played around with it.

fgheysels:
Thank you for picking this up.  I really look forward to be able to use this in a service in where I do not want a dependency on AspNet.
This structure is also more inline with other packages (SeriLog for instance, where they also have put the aspnet core related things in a dedicated package).

src/Arcus.Observability.Telemetry.AspNetCore/Arcus.Observability.Telemetry.AspNetCore.csproj
stijnmoreels: >  I assume that .NET 5 also depends on Microsoft.AspnetCore.App,
I think so too: https://github.com/dotnet/core/blob/master/release-notes/5.0/5.0.1/5.0.1.md

OK, so we should indeed check for `netstandard` so we can later move to .NET 5.

src/Arcus.Observability.Telemetry.AspNetCore/Extensions/ILoggerExtensions.cs
stijnmoreels: Resolving this, as I've made the `RequestFormat` `public`.

src/Arcus.Observability.Telemetry.AspNetCore/Arcus.Observability.Telemetry.AspNetCore.csproj
stijnmoreels: I also updated the correlation project bc it had the same conditional check there.

src/Arcus.Observability.Telemetry.Core/MessageFormats.cs
```diff
@@ -0,0 +1,21 @@
+﻿namespace Arcus.Observability.Telemetry.Core
+{
+    /// <summary>
+    /// Represents the publicly available message formats to track telemetry.
+    /// </summary>
+    public static class MessageFormats
```
tomkerkhove, line 6: Maybe consider moving all formats?

tomkerkhove (approved):

src/Arcus.Observability.Telemetry.Core/MessageFormats.cs
stijnmoreels: Yeah, maybe that's a better approach. More consistent too.

stijnmoreels: Will still use the 'old' message format for now to track events, will change it in #167 .

stijnmoreels: Done this, resolving.

src/Arcus.Observability.Telemetry.AspNetCore/Arcus.Observability.Telemetry.AspNetCore.csproj
fgheysels: Sorry for me to be a bit picky on this, but I'm not convinced that the changes that have been done now, are good.  

What frameworks are we supporting in Arcus ? 
I'd say that if we support .NET 5.0, .NET Core 3.x and .NET standard, that we should maybe explicitly test on those.

For instance :
```csharp
<ItemGroup Condition = "'$(TargetFramework)' = 'net5.0' or '$(TargetFramework)' = 'netcoreapp3.1'">
```
and 
```csharp
<ItemGroup Condition="'$(TargetFramework)' = 'netstandard2.0'>
```

These are assumptions based on [this](https://docs.microsoft.com/en-us/dotnet/standard/frameworks#supported-target-frameworks) article.

(Don't know if this works, 'cause I haven't tested it or played with it).

fgheysels: Related:  should we explicitly support .NET 5 ?  @tomkerkhove  ?

fgheysels: I was going to try to use the preview build on myget in a .NET 5 app to see what it gives, but it looks like there's no preview for observability available on myget ?

tomkerkhove: > Related: should we explicitly support .NET 5 ? @tomkerkhove ?

Nothing planned yet but feel free to open an issue since we should start supporting it



> I was going to try to use the preview build on myget in a .NET 5 app to see what it gives, but it looks like there's no preview for observability available on myget ?

You should see it though, see https://dev.azure.com/codit/Arcus/_build/results?buildId=50261&view=logs&jobId=02805f64-2152-5b7a-07d4-706789a8a527&j=02805f64-2152-5b7a-07d4-706789a8a527&t=79e516e3-a0d7-5428-bbee-4af2fc23326b

fgheysels: Ah, I had to check the 'preview' checkbox.

But, it's not clear to me which package I need to use now if I want to make use of this library when I have no dependency on AspNetCore.
These are the packages that are available:

image

I see that there's a Arcus.Observability.Telemetry.AspNetCore package, which I don't want to use.  This is the one that needs to be used in Api projects or other projects that depend on aspnet core.
The package Arcus.Observability.Telemetry.Serilog package seems to be an old one, so it isn't that one either I guess.

Is it ....Sinks.ApplicationInsights ?

Next to that: are the Telemetry.IoT and Telemetry.Sql packages enrichers ? If so, why aren't they called `Telemetry.Serilog.Enrichers.IoT` and `Telemetry.Serilog.Enrichers.Sql` ?  
Why do some packages have Serilog in their name, and others don't ?  It looks like all packages depend on Serilog ?

stijnmoreels: > But, it's not clear to me which package I need to use now if I want to make use of this library when I have no dependency on AspNetCore.

This PR fix is about removing the ASP.NET Core dependency from our general Core package. So that's the `Arcus.Observability.Telemetry.Core` package.

> I see that there's a Arcus.Observability.Telemetry.AspNetCore package, which I don't want to use. This is the one that needs to be used in Api projects or other projects that depend on aspnet core.

Yes, this package is about tracking dependencies directly from `HttpRequest`/`HttpResponse` types without extracting the valuable information yourself.

> The package Arcus.Observability.Telemetry.Serilog package seems to be an old one, so it isn't that one either I guess.

Yes that's an old one. This is now split into multiple package, each with a different purpose (enrichers, sinks...)

 > Is it ....Sinks.ApplicationInsights ?

This package is about the Application Insights implementation of a Serilog sink. Not directly related to telemetry tracking itself.

> Next to that: are the Telemetry.IoT and Telemetry.Sql packages enrichers ? 

No, these packages are about tracking dependencies with specific IoT, Sql, AspNetCore... types. The reason they are in separate packages is because they require additional NuGet packages and we don't want them in our general telemetry core library.

>  If so, why aren't they called Telemetry.Serilog.Enrichers.IoT and Telemetry.Serilog.Enrichers.Sql ?

As answered above, they're not enricher packages and are not related to anything about Serilog.

> Why do some packages have Serilog in their name, and others don't ?

The packages with 'Serilog' in their name is in indication that these libraries are extensions on Serilog (enrichers, sinks, ...). The packages without the 'Serilog' in their name is about telemetry tracking but has in fact nothing to do with Serilog directly. 

>  It looks like all packages depend on Serilog ?

No, that's no true. The packages without 'Serilog' in their name is about general telemetry tracking. It can be used outside an Serilog environment.

stijnmoreels: > Sorry for me to be a bit picky on this, but I'm not convinced that the changes that have been done now, are good.
> 
> What frameworks are we supporting in Arcus ?
> I'd say that if we support .NET 5.0, .NET Core 3.x and .NET standard, that we should maybe explicitly test on those.
> 
> For instance :
> 
> ```cs
> <ItemGroup Condition = "'$(TargetFramework)' = 'net5.0' or '$(TargetFramework)' = 'netcoreapp3.1'">
> ```
> 
> and
> 
> ```cs
> <ItemGroup Condition="'$(TargetFramework)' = 'netstandard2.0'>
> ```
> 
> These are assumptions based on [this](https://docs.microsoft.com/en-us/dotnet/standard/frameworks#supported-target-frameworks) article.
> 
> (Don't know if this works, 'cause I haven't tested it or played with it).

@tomkerkhove , I think it's better to take this up in another issue (maybe even on the general Arcus repo) so that we can discus that outside this issue. Since this PR is about removing the ASP.NET Core dependency and has as of now nothing to do (yet) with an upcoming .NET version.

fgheysels:
I just tried this preview out in a project where I'd like to use this functionality.  This project has no dependency on AspNet, so it would be an ideal candidate.
However, after using the packages `Arcus.Observability.Telemetry.Serilog.Sinks.ApplicationInsights` and `Arcus.Observability.Telemetry.Core`(which I do not need to reference explicitly, since it is already referenced by the ...Sinks.ApplicationInsights package, I'm still not able to use these assemblies in a docker container.    
I still get the error that the program cannot be started since the AspNet Framework is not present in my container:

```
It was not possible to find any compatible framework version
The framework 'Microsoft.AspNetCore.App', version '3.1.0' was not found.
  - No frameworks were found.

You can resolve the problem by installing the specified framework and/or SDK.
```

I don't need the AspNetCore framework inside my container, so I don't want to install it since I want to keep my container as small as possible.

After some digging around in this branch, I saw that the `Arcus.Observability.Correlation` [assembly still has a dependency on AspNetCore](https://github.com/stijnmoreels/arcus.observability/blob/d316b645b1652505f595dd2551f4b6f131405fb3/src/Arcus.Observability.Correlation/Arcus.Observability.Correlation.csproj#L20).
(`Arcus.Observability.Correlation` is indirectly used by `Arcus.Observability.Telemetry.Serilog.Sinks.ApplicationInsights` via `Arcus.Observability.Serilog.Enrichers`).

docs/preview/features/writing-different-telemetry-types.md
```diff
@@ -373,6 +373,16 @@ logger.LogMetric("Invoice Received", 133.37, telemetryContext);
...
+**Installation**
+
+If you want to track the `HttpRequest` and `HttpResponse` of an ASP.NET Core project, you'll have to install an additional package to include these ASP.NET Core dependencies:
+
+```shell
+PM > Install-Package Arcus.Observability.Telemetry.AspNetCore
+```
+
+**Example**
+
```
fgheysels, line 385: It would be nice to extend the documentation regarding application insights with a more extended code-sample.  Right now, some code is shown on how to log to AppInsights via Serilog, but it is not shown where you could put that code.

src/Arcus.Observability.Correlation/Arcus.Observability.Correlation.csproj
```diff
@@ -17,11 +17,11 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
   </PropertyGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.1'">
+  <ItemGroup Condition="'$(TargetFramework)' != 'netstandard2.0'">
     <FrameworkReference Include="Microsoft.AspNetCore.App" />
   </ItemGroup>
```
fgheysels, line 22: is the dependency to AspNetCore necessary here ?
The new assembly `Arcus.Obsservability.Telemetry.Core` has a dependency on this assembly (Correlation), so that means that it is still not possible to make use of the functionalities exposed by this library in projects that are hosted in containers that do not have the aspnet framework installed.  (See my other comment in the converstations tab).

fgheysels (changes requested):

src/Arcus.Observability.Telemetry.AspNetCore/Arcus.Observability.Telemetry.AspNetCore.csproj
fgheysels: I agree to have another look at it via another issue / PR.   Current implementation might be sufficient for now.

src/Arcus.Observability.Correlation/Arcus.Observability.Correlation.csproj
tomkerkhove: I don't think so as well

src/Arcus.Observability.Telemetry.Core/Arcus.Observability.Telemetry.Core.csproj
```diff
@@ -19,16 +19,8 @@
...
-    <FrameworkReference Include="Microsoft.AspNetCore.App" />
-  </ItemGroup>
-
-  <ItemGroup Condition="'$(TargetFramework)' != 'netcoreapp3.1'">
-    <PackageReference Include="Microsoft.AspNetCore.Http.Abstractions" Version="2.2.0" />
-    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.2.0" />
-  </ItemGroup>
-
   <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="3.1.8" />
```
tomkerkhove, line 23: If it was added with a condition here (https://github.com/arcus-azure/arcus.observability/pull/166/files#diff-94bb8bf830bf69c261860dcf4886e0c762d193d7bb8eda88b9bd7451618adce5L26), why are we always adding it now? Is it on purpose?

src/Arcus.Observability.Telemetry.AzureFunctions/Arcus.Observability.Telemetry.AzureFunctions.csproj
```diff
@@ -17,11 +17,11 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
   </PropertyGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.1'">
+  <ItemGroup Condition="'$(TargetFramework)' != 'netstandard2.0'">
     <FrameworkReference Include="Microsoft.AspNetCore.App" />
```
tomkerkhove, line 21: Think about this one, should we move it out or not as well or is it overkill given Azure Functions brings it in already @fgheysels ?

src/Arcus.Observability.Correlation/Arcus.Observability.Correlation.csproj
stijnmoreels: Ah, yes, it was previously required bc of the correlation middleware, but that's moved.

src/Arcus.Observability.Telemetry.Core/Arcus.Observability.Telemetry.Core.csproj
stijnmoreels: Not sure what you're getting at? Is it about the packages or the separate HTTP project?

src/Arcus.Observability.Correlation/Arcus.Observability.Correlation.csproj
stijnmoreels: I also remember a previous issue that we fixed on all the repo's where we were supposed to add the AspNetCore dependency to be used when we required netcoreapp3.1 dependencies. I guess when I see the comments here, that that issue proposal was wrong.

src/Arcus.Observability.Telemetry.AzureFunctions/Arcus.Observability.Telemetry.AzureFunctions.csproj
stijnmoreels: Maybe here, we should also use the 3.1.8 packages. Diff issue perhaps. It's already too much in this PR for my opinon.

src/Arcus.Observability.Correlation/Arcus.Observability.Correlation.csproj
tomkerkhove: I don't think we have to do that indeed, we should only add ASP.NET when we need it.

src/Arcus.Observability.Telemetry.Core/Arcus.Observability.Telemetry.Core.csproj
tomkerkhove: We had a condition for this one first and now we always add it. I'm just wondering if that was on purpose or not; that's all

stijnmoreels: It was on purpose. Apparently, the 3.1.8 version supports netstandard, but I'm still not sure if that means that we support netcoreapp2.2.

src/Arcus.Observability.Telemetry.AspNetCore/Arcus.Observability.Telemetry.AspNetCore.csproj
stijnmoreels: Created: https://github.com/arcus-azure/arcus/issues/106

src/Arcus.Observability.Correlation/Arcus.Observability.Correlation.csproj
fgheysels: Indeed, I think that this has already been discussed some time ago:  only add AspNetCore framework as a dependency when it's required.  And, I think that this is already done for most projects ?

src/Arcus.Observability.Telemetry.Core/Arcus.Observability.Telemetry.Core.csproj
tomkerkhove: We can drop 2.2 support if we have to

src/Arcus.Observability.Telemetry.AzureFunctions/Arcus.Observability.Telemetry.AzureFunctions.csproj
fgheysels: The dependency to AspNetCore.App should be removed here indeed.   AFAIK, an Azure Function has no need to use the AspnetCore framework.

The Microsoft.Extensions.Logging dependency should just be added, regardless if you're targetting netstandard2.0 or netcoreapp3.1.   (I'll leave it up to you guys to decide which version you should use.     I'd think the lowest  version possible ?  )

src/Arcus.Observability.Correlation/Arcus.Observability.Correlation.csproj
```diff
@@ -17,11 +17,12 @@
...
+  <ItemGroup Condition="'$(TargetFramework)' != 'netstandard2.0'">
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="3.1.8" />
+    <PackageReference Include="Microsoft.Extensions.Options" Version="3.1.8" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' != 'netcoreapp3.1'">
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="2.2.0" />
     <PackageReference Include="Microsoft.Extensions.Options" Version="2.2.0" />
   </ItemGroup>
```
fgheysels, line 28: Is it a problem if we just target the lowest version possible ?  2.2.0 targets .net standard I suppose, and .NET Core 3.1 also implements .NET standard, so that should just work.

fgheysels (changes requested):

src/Arcus.Observability.Telemetry.AzureFunctions/Arcus.Observability.Telemetry.AzureFunctions.csproj
stijnmoreels: Th telemetry core package already has a 3.1.8 version package for the logging. Maybe we should stick with that version in the telemetry Azure Functions package.

stijnmoreels: So, I fixed it's also in the telemetry core package changed. I saw that we didn't released this change with the 3.1.8 version everywhere; so we can change that.

src/Arcus.Observability.Telemetry.Core/Arcus.Observability.Telemetry.Core.csproj
fgheysels: I'd assume that .NET Core 2.2 also implements .NET Standard 2.

src/Arcus.Observability.Telemetry.AzureFunctions/Arcus.Observability.Telemetry.AzureFunctions.csproj
fgheysels: This also means that all projects that are using this Arcus library, will have to upgrade their version of the logging package if they're also explicitly using it.

src/Arcus.Observability.Correlation/Arcus.Observability.Correlation.csproj
```diff
@@ -17,11 +17,12 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
   </PropertyGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.1'">
-    <FrameworkReference Include="Microsoft.AspNetCore.App" />
+  <ItemGroup Condition="'$(TargetFramework)' != 'netstandard2.0'">
```
fgheysels, line 20: If the targetframework is netstandard1.0, this will also evaluate to true ....   Maybe I'm too much of a nitpicker :)
But ok, I guess we'll have another look at it via another PR.  (And the supported frameworks are now only netstandard2.0 and netcoreapp3.1, so then it's not a problem I guess.

fgheysels (approved):
Thanks for the work!

PR #170, opened 2021-01-05
Add guidance on where to configure the logger

### LINKED ISSUE #168: Improve sample in AppInsights documentation
It would be nice to extend the documentation regarding application insights with a more extended code-sample.  Right now, some code is shown on how to log to AppInsights via Serilog, but it is not shown where you could put that code.

I think it would be better for Arcus users to have some more context on that.

fgheysels:
Added some guidance in the Q&A section of the documentation on how to use / configure the logger.

Closes #168

### DIFF
--- a/docs/features/sinks/azure-application-insights.md
+++ b/docs/features/sinks/azure-application-insights.md
@@ -86,7 +86,7 @@ var host = Host.CreateDefaultBuilder()
                .Build();
 ```
 
-If the InstrumentationKey is stored as a secret in -for instance- Azure KeyVault, the `ISecretProvider` from [Arcus.Security](https://github.com/arcus-azure/arcus.security) can be used to retrieve the InstrumentationKey.
+If the InstrumentationKey is stored as a secret in -for instance- Azure Key Vault, the `ISecretProvider` from [Arcus Security](https://github.com/arcus-azure/arcus.security) can be used to retrieve the instrumentationKey.
 
 ```csharp
 using Serilog.Configuration;
--- a/docs/features/sinks/azure-application-insights.md
+++ b/docs/features/sinks/azure-application-insights.md
@@ -62,6 +62,10 @@ ILogger logger = loggerConfig.CreateLogger();
 Simply use the `UseSerilog` extension method on `IHostBuilder` which accepts an `Action<HostBuilderContext, IServiceProvider, LoggerConfiguration>`.  This Action gives you access to the configuration and the configured services:
 
 ```csharp
+using Serilog.Configuration;
+
+...
+
 var host = Host.CreateDefaultBuilder()
                .ConfigureAppConfiguration((context, configBuilder) =>
                {
@@ -73,6 +77,36 @@ var host = Host.CreateDefaultBuilder()
                               .WriteTo.Console();
 
                     string instrumentationKey = context.Configuration["ApplicationInsights:InstrumentationKey"];
+
+                    if (!String.IsNullOrWhiteSpace(instrumentationKey))
+                    {
+                        loggerConfiguration.WriteTo.AzureApplicationInsights(instrumentationKey, LogEventLevel.Information);
+                    }
+               })
+               .Build();
+```
+
+If the InstrumentationKey is stored as a secret in -for instance- Azure KeyVault, the `ISecretProvider` from [Arcus.Security](https://github.com/arcus-azure/arcus.security) can be used to retrieve the InstrumentationKey.
+
+```csharp
+using Serilog.Configuration;
+using Arcus.Security.Core;
+
+...
+
+var host = Host.CreateDefaultBuilder()
+               .ConfigureSecretStore((context, config, builder) =>
+               {
+                   // Configure the secretstore here
+               })
+               .UseSerilog((context, serviceProvider, loggerConfig) =>
+               {
+                    loggerConf.Enrich.FromLogContext()
+                              .WriteTo.Console();
+
+                    var secretProvider = serviceProvider.GetService<ISecretProvider>();
+
+                    var instrumentationKey = secretProvider.GetRawSecretAsync("ApplicationInsights:InstrumentationKey").GetAwaiter().GetResult();
 
                     if (!String.IsNullOrWhiteSpace(instrumentationKey))
                     {
--- a/docs/features/sinks/azure-application-insights.md
+++ b/docs/features/sinks/azure-application-insights.md
@@ -59,7 +59,9 @@ ILogger logger = loggerConfig.CreateLogger();
 
 ### Q: Where can I initialize the logger in an ASP.NET Core application or other hosted service?
 
-The Application Insights instrumentation key is typically available via the configuration or via an `ISecreteProvider`.  It's best if the Logger is created and assigned to the Serilog `Log.Logger` as soon as possible.  The easiest place to do that in an ASP.NET application  is in the `ConfigureServices` method of the `Program` class.
+The Azure Application Insights instrumentation key is typically available via the configuration or `ISecreteProvider`.  It's best if the logger is created and assigned to the Serilog `Log.Logger` as soon as possible.
+
+The easiest place to do that in an ASP.NET application is in the `ConfigureServices` method of the `Program` class.
 
 Another place where the logger can be initialized is in the `ConfigureAppConfiguration` method of the `IHostBuilder`:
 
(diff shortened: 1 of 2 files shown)


docs/features/sinks/azure-application-insights.md
```diff
@@ -57,4 +57,66 @@ if(string.IsNullOrEmpty(key) == false)
...
+
+```csharp
+var host = Host.CreateDefaultBuilder()
+               .ConfigureSecretStore((context, config, builder) =>
+               {
+                   // Configure the secretstore here
+               })
+               .ConfigureServices((context, services) =>
+               {
+                   var serviceProvider = services.BuildServiceProvider();
```
stijnmoreels, line 101: Building a service provider yourself is not recommended and is probably not a good idea to use as an example to newcomers.
If you need the service provider, look at this example in our templates: https://github.com/arcus-azure/arcus.templates/blob/master/src/Arcus.Templates.WebApi/Startup.cs#L306

And see the recommendation here: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-3.1#recommendations

docs/features/sinks/azure-application-insights.md
```diff
@@ -57,4 +57,66 @@ if(string.IsNullOrEmpty(key) == false)
...
+                    {
+                        logConfiguration.WriteTo.AzureApplicationInsights(appInsightsInstrumentationKey);
+                    }
+
+                    Log.Logger = logConfiguration.CreateLogger();
+               })
+               .UseSerilog()
+               .Build();
+```
+
```
stijnmoreels, line 90: Maybe a link to the Arcus Security library is useful here.

stijnmoreels (changes requested):
Thank you for this. Really helpful to have someone outside to write this.

docs/features/sinks/azure-application-insights.md
```diff
@@ -57,4 +57,66 @@ if(string.IsNullOrEmpty(key) == false)
 ILogger logger = loggerConfig.CreateLogger();
 ```
 
+### Q: Where can I initialize the logger in an ASP.NET Core application or other hosted service?
+
+The Application Insights instrumentation key is typically available via the configuration or via an `ISecreteProvider`.  It's best if the Logger is created and assigned to the Serilog `Log.Logger` as soon as possible.  The easiest place to do that in an ASP.NET application  is in the `ConfigureServices` method of the `Program` class.
```
tomkerkhove, line 62:
```suggestion
The Azure Application Insights instrumentation key is typically available via the configuration or `ISecreteProvider`.  It's best if the logger is created and assigned to the Serilog `Log.Logger` as soon as possible.

The easiest place to do that in an ASP.NET application is in the `ConfigureServices` method of the `Program` class.
```

However, we actually do it as part of UseSerilog: https://github.com/arcus-azure/arcus.templates/blob/master/src/Arcus.Templates.ServiceBus.Queue/Program.cs#L74 but use a start-up logger to catch all logs before that: https://github.com/arcus-azure/arcus.templates/blob/master/src/Arcus.Templates.ServiceBus.Queue/Program.cs#L31-L34

tomkerkhove: With the current proposal you can't use anything from the secret store

docs/features/sinks/azure-application-insights.md
fgheysels: Thanks for the link, interesting!

So, what do you suggest then ?  That we configure the Logger in the `ConfigureWebHost` method for ASP.NET applications ?
And for other 'hosted' applications, that we do this just after the host has been build, like this ?

```csharp
var host = Host.CreateDefaultBuilder(args)
                       .ConfigureServices( ... )
                       .ConfigureSecretStore ( .... )  
                       .UseConsoleLifetime()
                      .UseSeriLog()
                       .Build();

var secretProvider = host.Services.GetService<ISecretProvider>();

var instrumentationKey = secretProvider.GetRawSecretAsync("ApplicationInsights:InstrumentationKey").GetAwaiter().GetResult();
var logConfiguration = new LoggerConfiguration()
                              .Enrich.FromLogContext()
                              .WriteTo.Console();
if (!String.IsNullOrWhiteSpace(appInsightsInstrumentationKey))
{
    logConfiguration.WriteTo.AzureApplicationInsights(appInsightsInstrumentationKey);
}

Log.Logger = logConfiguration.CreateLogger();

 await host.RunAsync();
```

docs/features/sinks/azure-application-insights.md
fgheysels: Oh, so you have an overload UseSeriLog method which allows you to configure the logger ?  That's actually even better.

fgheysels: > With the current proposal you can't use anything from the secret store

Can you please elaborate ?

docs/features/sinks/azure-application-insights.md
stijnmoreels: Hmm, I think in the Arcus templates is a good place to look. We configure there actually the logger twice; once before the actual creation of the `IHost` so we definitely have a logger if the setup fails and we want to log some critical information; and while the the host is created so we can make use of the `IServiceProvider`. 
But I think the way it's set up; the instrumation key is mostly picked from the environment variables, via the configuration. The secret store is not something we have/could use easily here I think (like Tom said below).

docs/features/sinks/azure-application-insights.md
tomkerkhove: No, more because it's all part of `ConfigureAppConfiguration`

docs/features/sinks/azure-application-insights.md
fgheysels: I've changed this so that the logger is configured in the UseSerilog extension method.
I've only added an example where the instrumentation key is retrieved from the configuration.   However, the UseSerilog extension method exposes an IServiceProvider as well, so I think we'll be able to retrieve the key from keyvault there as well.   Will need to try it first before I put it in the docs.

stijnmoreels (approved):
Great addition! You can choose to use indeed the `IServiceProvider` in the `.UseSerilog` if you want. Can be a good alternative for people. Too bad we don't have a sync variant for the `ISecretProvider.GetRawSecretAsync`.

fgheysels:
Still need to update the preview docs.

docs/features/sinks/azure-application-insights.md
```diff
@@ -57,4 +57,29 @@ if(string.IsNullOrEmpty(key) == false)
 ILogger logger = loggerConfig.CreateLogger();
 ```
 
+### Q: Where can I initialize the logger in an ASP.NET Core application or other hosted service?
+
+Simply use the `UseSerilog` extension method on `IHostBuilder` which accepts an `Action<HostBuilderContext, IServiceProvider, LoggerConfiguration>`.  This Action gives you access to the configuration and the configured services:
+
+```csharp
+var host = Host.CreateDefaultBuilder()
```
tomkerkhove, line 65: Should we add the required using statement for our sink?

tomkerkhove (approved):
LGTM, one suggestion but up to you.

docs/features/sinks/azure-application-insights.md
fgheysels: That might be helpful.  Added it.

fgheysels:
> Still need to update the preview docs.

This is done as well.

docs/features/sinks/azure-application-insights.md
fgheysels: Added a sample where Arcus Security is used + added a link to the Arcus.Security GH project

fgheysels:
> Still need to update the preview docs.

done.

docs/features/sinks/azure-application-insights.md
```diff
@@ -57,4 +57,63 @@ if(string.IsNullOrEmpty(key) == false)
...
+
+                    if (!String.IsNullOrWhiteSpace(instrumentationKey))
+                    {
+                        loggerConfiguration.WriteTo.AzureApplicationInsights(instrumentationKey, LogEventLevel.Information);
+                    }
+               })
+               .Build();
+```
+
+If the InstrumentationKey is stored as a secret in -for instance- Azure KeyVault, the `ISecretProvider` from [Arcus.Security](https://github.com/arcus-azure/arcus.security) can be used to retrieve the InstrumentationKey.
```
tomkerkhove, line 89:
```suggestion
If the InstrumentationKey is stored as a secret in -for instance- Azure Key Vault, the `ISecretProvider` from [Arcus Security](https://github.com/arcus-azure/arcus.security) can be used to retrieve the instrumentationKey.
```

tomkerkhove (approved):

tomkerkhove:
Thanks @fgheysels!

PR #181, opened 2021-02-24
Log starttime of dependencies up to the millisecond

### LINKED ISSUE #180: tracked dependencies do not show up correctly in the Application Insights timeline
**Describe the bug**
When I use the Arcus.Observability `LogDependency` extension method on `ILogger`, I've noticed that my tracked dependencies do not show up correctly in the Application Insights timeline:

image
As you can see in the above screenshot, it looks like that my dependency to Timeseries Insights occurred before the initial request to my API was received, which is actually impossible, as my API is calling TSI; so it is not possible that this dependency occurs before the initial request on the API was received.

After some digging around, I have found that it all comes down to the precision of the timestamp that is being logged.  The timestamp of my API request is this:

```
2021-02-24T13:33:26.0349398Z
```

Whereas the timestamp of my tracked TSI dependency is this:

```
2021-02-24T13:33:26Z
```
As you can see, the dependency log's timestamp is only precise to the second, whereas the timestamp of the incoming API request is precise up to the millisecond.
The timestamp of the TSI dependency log will be considered to be 13:33:26.000Z which is then indeed before 13:33:26.0349Z

fgheysels:
When dependencies are tracked, the starttime of the dependency is converted to a string using the InvariantCulture format provider.  However, this provider only converts the datetime to a string with a precision up to the second.
There doesn't seem to be any built-in format provider that converts dates to strings with millisecond precision, so we've specified our own.

closes #180

### DIFF
--- a/src/Arcus.Observability.Telemetry.AspNetCore/Extensions/ILoggerExtensions.cs
+++ b/src/Arcus.Observability.Telemetry.AspNetCore/Extensions/ILoggerExtensions.cs
@@ -12,8 +12,6 @@ namespace Microsoft.Extensions.Logging
     /// </summary>
     public static class ILoggerExtensions
     {
-        private static string InvariantDateTimeWithMillisecondsFormat = "yyyy-MM-ddTHH:mm:ss.fffffff zzz";
-
         /// <summary>
         ///     Logs an HTTP request
         /// </summary>
@@ -55,7 +53,7 @@ namespace Microsoft.Extensions.Logging
             PathString resourcePath = request.Path;
             var host = $"{request.Scheme}://{request.Host}";
 
-            logger.LogWarning(MessageFormats.RequestFormat, request.Method, host, resourcePath, responseStatusCode, duration, DateTimeOffset.UtcNow.ToString(InvariantDateTimeWithMillisecondsFormat), context);
+            logger.LogWarning(MessageFormats.RequestFormat, request.Method, host, resourcePath, responseStatusCode, duration, DateTimeOffset.UtcNow.ToString(FormatSpecifiers.InvariantTimestampFormat), context);
         }
     }
 }
--- a/src/Arcus.Observability.Telemetry.AspNetCore/Extensions/ILoggerExtensions.cs
+++ b/src/Arcus.Observability.Telemetry.AspNetCore/Extensions/ILoggerExtensions.cs
@@ -12,7 +12,7 @@ namespace Microsoft.Extensions.Logging
     /// </summary>
     public static class ILoggerExtensions
     {
-        private static string InvariantDateTimeWithMillisecondsFormat = "yyyy-MM-ddTHH:mm:ss.fff zzz";
+        private static string InvariantDateTimeWithMillisecondsFormat = "yyyy-MM-ddTHH:mm:ss.fffffff zzz";
 
         /// <summary>
         ///     Logs an HTTP request
--- a/src/Arcus.Observability.Telemetry.AspNetCore/Extensions/ILoggerExtensions.cs
+++ b/src/Arcus.Observability.Telemetry.AspNetCore/Extensions/ILoggerExtensions.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Globalization;
 using Arcus.Observability.Telemetry.Core;
 using GuardNet;
 using Microsoft.AspNetCore.Http;
@@ -10,6 +9,8 @@ namespace Microsoft.Extensions.Logging
 {
     public static class ILoggerExtensions
     {
+        private static string InvariantDateTimeWithMillisecondsFormat = "yyyy-MM-ddTHH:mm:ss.fff zzz";
+
         // <summary>
         ///     Logs an HTTP request
         /// </summary>
@@ -51,7 +52,7 @@ namespace Microsoft.Extensions.Logging
             PathString resourcePath = request.Path;
             var host = $"{request.Scheme}://{request.Host}";
 
-            logger.LogWarning(MessageFormats.RequestFormat, request.Method, host, resourcePath, responseStatusCode, duration, DateTimeOffset.UtcNow.ToString(CultureInfo.InvariantCulture), context);
+            logger.LogWarning(MessageFormats.RequestFormat, request.Method, host, resourcePath, responseStatusCode, duration, DateTimeOffset.UtcNow.ToString(InvariantDateTimeWithMillisecondsFormat), context);
         }
     }
 }
--- a/src/Arcus.Observability.Telemetry.AspNetCore/Extensions/ILoggerExtensions.cs
+++ b/src/Arcus.Observability.Telemetry.AspNetCore/Extensions/ILoggerExtensions.cs
@@ -7,11 +7,14 @@ using Microsoft.AspNetCore.Http;
 // ReSharper disable once CheckNamespace
 namespace Microsoft.Extensions.Logging
 {
+    /// <summary>
+    /// Telemetry extensions on the <see cref="ILogger"/> instance to write Application Insights compatible log messages.
+    /// </summary>
     public static class ILoggerExtensions
     {
         private static string InvariantDateTimeWithMillisecondsFormat = "yyyy-MM-ddTHH:mm:ss.fff zzz";
 
-        // <summary>
+        /// <summary>
         ///     Logs an HTTP request
         /// </summary>
         /// <param name="logger">Logger to use</param>
--- a/src/Arcus.Observability.Tests.Integration/Serilog/Sinks/ApplicationInsights/ApplicationInsightsSinkTests.cs
+++ b/src/Arcus.Observability.Tests.Integration/Serilog/Sinks/ApplicationInsights/ApplicationInsightsSinkTests.cs
@@ -65,7 +65,7 @@ namespace Arcus.Observability.Tests.Integration.Serilog.Sinks.ApplicationInsight
 
         protected async Task RetryAssertUntilTelemetryShouldBeAvailableAsync(Func<Task> assertion)
         {
-            await RetryAssertUntilTelemetryShouldBeAvailableAsync(assertion, timeout: TimeSpan.FromMinutes(7));
+            await RetryAssertUntilTelemetryShouldBeAvailableAsync(assertion, timeout: TimeSpan.FromMinutes(12));
         }
 
         protected async Task RetryAssertUntilTelemetryShouldBeAvailableAsync(Func<Task> assertion, TimeSpan timeout)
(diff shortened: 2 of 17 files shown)


tomkerkhove (approved):

fgheysels:
Don't merge this one yet please (when the tests succeed :) ), I'd like to test / validate this by integrating the preview build in a project.

src/Arcus.Observability.Tests.Integration/Serilog/Sinks/ApplicationInsights/AzureSearchDependencyTests.cs
```diff
@@ -30,7 +30,7 @@ public async Task LogAzureSearchDependency_SinksToApplicationInsights_ResultsInA
                 ILogger logger = loggerFactory.CreateLogger<ApplicationInsightsSinkTests>();
 
                 bool isSuccessful = BogusGenerator.PickRandom(true, false);
-                DateTimeOffset startTime = BogusGenerator.Date.RecentOffset(days: 0);
```
tomkerkhove, line 33: Is there a reason why we are changing this?

tomkerkhove (changes requested):
We'll need to update our sink as well:
- Our dependency telemetry converter reads it here: [code](https://github.com/arcus-azure/arcus.observability/blob/master/src/Arcus.Observability.Telemetry.Serilog.Sinks.ApplicationInsights/Converters/DependencyTelemetryConverter.cs#L25)
- That uses  GetAsDateTimeOffset here: [code](https://github.com/arcus-azure/arcus.observability/blob/4115c3d94785b55e388282e1441541f0d4053dcf/src/Arcus.Observability.Telemetry.Serilog.Enrichers/Extensions/LogEventPropertyValueExtensions.cs#L117)

src/Arcus.Observability.Tests.Integration/Serilog/Sinks/ApplicationInsights/AzureSearchDependencyTests.cs
fgheysels: For some reason, a lot of integration tests were failing.  It seemed like they were timing out because the dependencies that were written to AppInsights were not retrieved in time.   Sometimes I could reproduce it on my machine, sometimes not.
Then I decided to see what happens if I only retrieve the dependency-logs from AppInsights that occured in the last x minutes, by specifying a timespan filter.  Then, it looks like everything succeeds again.

Offcourse, to do this, I can't rely on random generated timestamps for the events.  I also suspect, but I have no proof for it, that writing a lot of things to appinsights with out-of-order timestamps might hurt AppInsights performance, as I would thing that AppInsights is timeseries based so some re-ordering might happen in appinisghts if you write out of order, but again, I only suspect that.
IMHO, it's also not a big deal to use the current datetime as the timestamp for the event that we're writing; it's actually as random as picking just another time for today's date.

tomkerkhove: I think this is related to the fact that we have not updated the application insights sink yet which is not respecting the milliseconds.

Can you test the same after making the changes and also with the original recent offset please?

fgheysels: I can't test it using the preview build in my project.  For some reason, it keeps complaining that Arcus.Observability.Correlation when I use the preview build in my project.

Even if I explicitly add all Arcus.Observability projects to my project, it keeps on complaining.
I also have a dependency on Arcus.WEbApi.Logging.Core; maybe they don't play well together ?

```
Could not load file or assembly 'Arcus.Observability.Correlation, Version=2.0.0.0, Culture=neutral, PublicKeyToken=null'. 
The system cannot find the file specified.
```

fgheysels: If I remove the Arcus.WebApi.Logging package from my project, I don't have the AssemblyLoadException anymore.

fgheysels: > I think this is related to the fact that we have not updated the application insights sink yet which is not respecting the milliseconds.
> 
> Can you test the same after making the changes and also with the original recent offset please?

I've tested the current preview build in my project, and the timestamp of the dependency is correct up to the millisecond.
However, I see that the timestamps of all other traces in AppInsights have an even higher precision (it's in ticks).  Think we should also go to that precision.  I'll do some updates here.

fgheysels: > I think this is related to the fact that we have not updated the application insights sink yet which is not respecting the milliseconds.

There are tests available which test if the date that is transformed by the appinsights converter is set correct.

src/Arcus.Observability.Telemetry.AspNetCore/Extensions/ILoggerExtensions.cs
```diff
@@ -1,16 +1,20 @@
...
 // ReSharper disable once CheckNamespace
 namespace Microsoft.Extensions.Logging
 {
+    /// <summary>
+    /// Telemetry extensions on the <see cref="ILogger"/> instance to write Application Insights compatible log messages.
+    /// </summary>
     public static class ILoggerExtensions
     {
-        // <summary>
+        private static string InvariantDateTimeWithMillisecondsFormat = "yyyy-MM-ddTHH:mm:ss.fffffff zzz";
```
tomkerkhove, line 15: Since we use it multiple times we should maybe move it to a shared place?

src/Arcus.Observability.Tests.Integration/Serilog/Sinks/ApplicationInsights/ApplicationInsightsSinkTests.cs
```diff
@@ -65,7 +65,7 @@ protected ILoggerFactory CreateLoggerFactory(Action<LoggerConfiguration> configu
 
         protected async Task RetryAssertUntilTelemetryShouldBeAvailableAsync(Func<Task> assertion)
```
tomkerkhove, line 66: Can we revert these changes?

tomkerkhove (approved):

src/Arcus.Observability.Telemetry.AspNetCore/Extensions/ILoggerExtensions.cs
fgheysels: Yes, I was thinking of that as well.   Problem is where to put it ?  I initially thought to put it in Telemetry.Core, but Telemetry.AspNetCore has no dependency on Telemetry.Core, and I thought it would maybe be overkill to create that dependency, but we can do that

tomkerkhove: Let's do that!

src/Arcus.Observability.Tests.Integration/Serilog/Sinks/ApplicationInsights/ApplicationInsightsSinkTests.cs
fgheysels: Yes we can; I just thought it was overkill to query AppInsights every second.

tomkerkhove: This seems like a long wait though?
https://github.com/arcus-azure/arcus.observability/pull/181/files/6c7cfe05d76513b14edd981cb8fe4602432ffbe1#diff-27f08c3b8059324291530235f56cd9bb92ef433ef8842cd03175f58a2213f06bR68

src/Arcus.Observability.Telemetry.AspNetCore/Extensions/ILoggerExtensions.cs
fgheysels: My bad, the dependency was already there.

fgheysels:
> We'll need to update our sink as well:
> 
> * Our dependency telemetry converter reads it here: [code](https://github.com/arcus-azure/arcus.observability/blob/master/src/Arcus.Observability.Telemetry.Serilog.Sinks.ApplicationInsights/Converters/DependencyTelemetryConverter.cs#L25)
> * That uses  GetAsDateTimeOffset here: [code](https://github.com/arcus-azure/arcus.observability/blob/4115c3d94785b55e388282e1441541f0d4053dcf/src/Arcus.Observability.Telemetry.Serilog.Enrichers/Extensions/LogEventPropertyValueExtensions.cs#L117)

I don't think we need to change anything to the DependencyTelemetryConverter.  When you feed a string representation of a datetimeoffset with ticks, they are converted correctly.  Lets see if I can add an additional test for it.

src/Arcus.Observability.Tests.Integration/Serilog/Sinks/ApplicationInsights/ApplicationInsightsSinkTests.cs
fgheysels: That's the max timeout.  So if it succeeds before that time, it won't wait that long

src/Arcus.Observability.Telemetry.AspNetCore/Extensions/ILoggerExtensions.cs
fgheysels: Done.

fgheysels:
> > We'll need to update our sink as well:
> > 
> > * Our dependency telemetry converter reads it here: [code](https://github.com/arcus-azure/arcus.observability/blob/master/src/Arcus.Observability.Telemetry.Serilog.Sinks.ApplicationInsights/Converters/DependencyTelemetryConverter.cs#L25)
> > * That uses  GetAsDateTimeOffset here: [code](https://github.com/arcus-azure/arcus.observability/blob/4115c3d94785b55e388282e1441541f0d4053dcf/src/Arcus.Observability.Telemetry.Serilog.Enrichers/Extensions/LogEventPropertyValueExtensions.cs#L117)
> 
> I don't think we need to change anything to the DependencyTelemetryConverter. When you feed a string representation of a datetimeoffset with ticks, they are converted correctly. Lets see if I can add an additional test for it.

Added an additional test, but I'm not that satisfied by that test actually.  Don't know if we can write it in another way

src/Arcus.Observability.Tests.Integration/Serilog/Sinks/ApplicationInsights/ApplicationInsightsSinkTests.cs
fgheysels: Put it back to the original (7 minutes).  Kept the 3 seconds delay though instead of the 1 sec delay between app insights queries

fgheysels:
> Don't merge this one yet please (when the tests succeed :) ), I'd like to test / validate this by integrating the preview build in a project.

Tested it, and it is ok (although I had to remove the WebApi.Logging package from my project to be able to test the preview nuget package from myget).

PR #197, opened 2021-03-29
fix: small typo in use-with-dotnet-and-functions.md

MassimoC:
small typo

### DIFF
--- a/docs/v2.1/guidance/use-with-dotnet-and-functions.md
+++ b/docs/v2.1/guidance/use-with-dotnet-and-functions.md
@@ -62,7 +62,7 @@ namespace Arcus.Samples.AzureFunction
 }
 ```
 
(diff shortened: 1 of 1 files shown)


docs/v2.1/guidance/use-with-dotnet-and-functions.md
```diff
@@ -62,7 +62,7 @@ namespace Arcus.Samples.AzureFunction
 }
 ```
 
-> :bulb: Not that we are using `ClearProvidersExceptFunctionProviders` instead of `ClearProviders` given Azure Functions requires some logging providers to be available.
+> :bulb: Note that we are using `ClearProvidersExceptFunctionProviders` instead of `ClearProviders` given Azure Functions requires some logging providers to be available.
```
stijnmoreels, line 65: Thanks! We'll have to check the other versions too, if this exists there also.

stijnmoreels (approved):

Issue #165, opened 2020-12-16
Provide dependency name for all dependencies


tomkerkhove:
**Is your feature request related to a problem? Please describe.**
Provide dependency name for all dependencies as it's not always tracked and is listed as `<empty>` as a result in Application Insights.

tomkerkhove:
I'll do a sweep of this and draft a list of expectations.

tomkerkhove:
This needs to be changed in `OperationContextConverter` and related to `telemetryEntry.Context.Operation.Name`.

However, this should be the name of the full operation that is being done, not just a dependency.

tomkerkhove:
https://docs.microsoft.com/en-gb/azure/azure-monitor/app/api-custom-events-metrics#operation-context

tomkerkhove:
This is good doc on how it should work: https://docs.microsoft.com/en-us/azure/azure-monitor/app/correlation#data-model-for-telemetry-correlation

tomkerkhove:
This is because `telemetryEntry.Context.Operation.Name` is empty:
![image](https://user-images.githubusercontent.com/4345663/128021510-7972791d-8480-412b-bcb4-0f1bb999522e.png)

tomkerkhove:
This dirty test verifies this hypothesis:
```csharp
public void EnrichWithCorrelationInfo<TEntry>(TEntry telemetryEntry) where TEntry : ITelemetry, ISupportProperties
        {
            if (telemetryEntry?.Context?.Operation == null)
            {
                return;
            }

            if (telemetryEntry.Properties.TryGetValue(ContextProperties.Correlation.OperationId, out string operationId))
            {
                telemetryEntry.Context.Operation.Id = operationId;
            }

            if (telemetryEntry.Properties.TryGetValue(ContextProperties.Correlation.OperationParentId, out string operationParentId))
            {
                telemetryEntry.Context.Operation.ParentId = operationParentId;
            }

            if (telemetryEntry is RequestTelemetry f)
            {
                f.Context.Operation.Name = f.Name;
            }

            if (telemetryEntry is DependencyTelemetry d)
            {
                d.Context.Operation.Name = d.Name;
            }
        }
```

stijnmoreels:
@pim-simons will take a first look at this to get a more feeling with the Arcus code. I'll help were necessary.

PR #207, opened 2021-04-30
feat: log telemetry as structured value models for Serilog formatting

### LINKED ISSUE #205: ILogger extension methods could add the Arcus additional meta-data in a structured way
I like the functionality that Arcus provides to easily log custom events , metrics, etc... very much.  
As far as I know, this functionality is made possible by adding additional meta-information to the message that is being logged, and the ApplicationInsights sink parses the log-message and retrieves that information.   This works, and this is great.

However, we have also services that are not logging to Application Insights directly.  Those services are running in containers in a Kubernetes cluster (on prem), and hence, the logs are written to the stdout & stderr of the containers.

We are using Azure Arc, and are transferring the stdout/stderr container logs via Azure Arc to Log Analytics.  We would like to work with those logs by executing KQL queries on the Log Analytics workspace.   The problem is, is that those logs are not in a structured format.   To be able to have them in a structured format, it would be great if the Arcus extension method could write the additional meta-information in a structured way in the log message.

I would suggest that the Arcus extension method do something like this.  (Warning, non compiled c# code; it's just to give you an idea on what could be possible):

```csharp
public static void LogEvent( this ILogger logger, string name, Dictionary<string,object> context )
{
     var entry = new LogEntry
     {
          Type = "Event",
          Message = name,
          Context = context;
     };
     logger.LogWarning("{@entry}", entry);
}

public class LogEntry
{
    public string LogType {get;set;}
    public string Message {get;set;}
    public Dictionary<string,object> Context {get;set;}
}
```

When you then use Serilogs' ConsoleLogger with this outputtemplate:

```csharp
loggerConfig.MinimumLevel.Information()
    .WriteTo.Console(
        outputTemplate: "{{ \"ts\": {Timestamp:yyyy-MM-ddTHH:mm:ss.zzz}, \"level\": \"{Level}\", \"message\": \"{Message:lj}\", \"properties\": {Properties:j} }{NewLine}",
        formatProvider: CultureInfo.InvariantCulture);
```

You get the output like this:

```
{ "ts": 2021-04-22T12:07:36+02:00, "level": "Warning", "message": "{"Type": "Event", "Message": "SomeEvent", "Properties": {"Item": "foo", "Error": "bar"}, "$type": "LogEntry"}", "properties": {"SourceContext": "ArcusObservabilityConsoleTest.TestClass"} }
```

So as you can see, the Message now contains all the requested information in a structured piece of json (note the `lj` formatter on `Message` in the template, so I guess would make this also easier for the AppInsights sink to retrieve the required information.  (Instead of doing some string parsing, you can deserialize the message to the `LogEntry` class and retrieve the information in a more typed way).

If Arcus could add the meta-information in this way, then I can have this as well in a structured way in Log Analytics :) 

More information on the Serilog specifics can be found [here](https://github.com/serilog/serilog/wiki/Formatting-Output).

stijnmoreels:
Change the way we log telemetry output by creating separate models that gets logged in one go without the 'sentence-like' formatting; which is not serializable.

Closes #205

### DIFF
--- a/docs/preview/features/writing-different-telemetry-types.md
+++ b/docs/preview/features/writing-different-telemetry-types.md
@@ -64,7 +64,7 @@ durationMeasurement.Start();
 var startTime = DateTimeOffset.UtcNow;
 
 logger.LogBlobStorageDependency(accountName: "multimedia", containerName: "images", isSuccessful: true, startTime, durationMeasurement.Elapsed);
-// Output: "Dependency DependencyLogTarget {DependencyType: Azure blob multimedia named images in 00:00:00.2521801 at 03/23/2020 09:56:31 +00:00 (Successful: True - Context: )"
+// Output: Dependency {"DependencyType": "Azure blob",  "DependencyName": "images", "TargetName": "multimedia", "Duration": "00:00:00.2521801", "StartTime": "03/23/2020 09:56:31 +00:00", "IsSuccessful": true, "Context": {}}
 ```
 
 ### Measuring Azure Cosmos DB dependencies
@@ -85,7 +85,7 @@ durationMeasurement.Start();
 var startTime = DateTimeOffset.UtcNow;
 
 logger.LogCosmosSqlDependency(accountName: "administration", database: "docs", container: "purchases", isSuccessful: true, startTime: startTime, duration: durationMeasurement.Elapsed);
-// Output: "Dependency Azure DocumentDB docs/purchases named administration in 00:00:00.2521801 at 03/23/2020 09:56:31 +00:00 (Successful: True - Context: )"
+// Output: Dependency {"DependencyType": "Azure DocumentDB", "DependencyData": "docs/purchases", "TargetName": "administration", "Duration": "00:00:00.2521801", "StartTime": "03/23/2020 09:56:31 +00:00", "IsSuccessful" true, "Context": {}}
 ```
 
 ### Measuring Azure Event Hubs dependencies
@@ -104,7 +104,7 @@ durationMeasurement.Start();
 var startTime = DateTimeOffset.UtcNow;
 
 logger.LogEventHubsDependency(namespaceName: "be.sensors.contoso", eventHubName: "temperature", isSuccessful: true, startTime: startTime, duration: durationMeasurement.Elapsed);
-// Output: "Dependency Azure Event Hubs be.sensors.contoso named temerature in 00:00:00.2521801 at 03/23/2020 09:56:31 +00:00 (Successful: True - Context: )"
+// Output: "Dependency {"DependencyType": "Azure Event Hubs", "DependencyData": "be.sensors.contoso", "TargetName": "temerature", "Duration": "00:00:00.2521801", "StartTime": "03/23/2020 09:56:31 +00:00", "IsSuccessful": true, "Context": {}}
 ```
 
 ### Measuring Azure IoT Hub dependencies
@@ -125,7 +125,7 @@ durationMeasurement.Start();
 var startTime = DateTimeOffset.UtcNow;
 
 logger.LogIotHubDependency(iotHubName: "sensors", isSuccessful: true, startTime: startTime, duration: durationMeasurement.Elapsed);
-// Output: "Dependency Azure IoT Hub named sensors in 00:00:00.2521801 at 03/23/2020 09:56:31 +00:00 (Successful: True - Context: )"
+// Output: Dependency {"DependencyType": "Azure IoT Hub", "TargetName": "sensors", "Duration": "00:00:00.2521801", "StartTime": "03/23/2020 09:56:31 +00:00", "IsSuccessful": true, "Context": {}}
 ```
 
 Or, alternatively you can pass allong the IoT connection string itself so the host name will be selected for you.
@@ -152,7 +152,7 @@ durationMeasurement.Start();
 var startTime = DateTimeOffset.UtcNow;
 
 logger.LogIotHubDependency(iotHubConnectionString: "Hostname=sensors;", isSuccessful: true, startTime: startTime, duration: durationMeasurement.Elapsed);
-// Output: "Dependency Azure IoT Hub named sensors in 00:00:00.2521801 at 03/23/2020 09:56:31 +00:00 (Successful: True - Context: )"
+// Output: Dependency {"DependencyType": "Azure IoT Hub", "TargetName": "sensors", "Duration": "00:00:00.2521801", "StartTime": "03/23/2020 09:56:31 +00:00", "IsSuccessful": true, "Context": {}}
 ```
 
 ### Measuring Azure Key Vault dependencies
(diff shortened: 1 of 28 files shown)


stijnmoreels:
Q: do we need to update the docs too? As in: should we change the log outputs in the https://observability.arcus-azure.net/features/writing-different-telemetry-types to this structured way?

docs/preview/features/writing-different-telemetry-types.md
```diff
@@ -64,7 +64,7 @@ durationMeasurement.Start();
 var startTime = DateTimeOffset.UtcNow;
 
 logger.LogBlobStorageDependency(accountName: "multimedia", containerName: "images", isSuccessful: true, startTime, durationMeasurement.Elapsed);
-// Output: "Dependency Azure blob multimedia named images in 00:00:00.2521801 at 03/23/2020 09:56:31 +00:00 (Successful: True - Context: )"
+// Output: "Dependency DependencyLogTarget {DependencyType: Azure blob multimedia named images in 00:00:00.2521801 at 03/23/2020 09:56:31 +00:00 (Successful: True - Context: )"
```
stijnmoreels, line 67: Q: do we need to update the docs too? As in: should we change the log outputs in the https://observability.arcus-azure.net/features/writing-different-telemetry-types to this structured way?

tomkerkhove: Yes please!

tomkerkhove: But the current change doesn't look correct though

stijnmoreels: Yes, I know, I was starting but stopped 😅 since I didn't know if I should change these too. It can be changed though the output template in Serilog and wasn't sure if it's a good idea to have a fixed representation of this here; because the way it's set up, you can still have the same output if you use the general `ToString` on our log models (and not use this Serilog decomposing formatting).

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
```diff
@@ -229,8 +229,16 @@ public static class ILoggerExtensions
             
             context = context ?? new Dictionary<string, object>();
 
-            logger.LogWarning(
-                DependencyFormat, dependencyType, dependencyData, targetName, duration, startTime.ToString(FormatSpecifiers.InvariantTimestampFormat), isSuccessful, context);
+            logger.LogWarning(DependencyFormat, new DependencyLogEntry(
+                dependencyType, 
+                dependencyName: null,
```
fgheysels, line 234: Why is dependencyName null ?

src/Arcus.Observability.Telemetry.Core/Logging/DependencyLogEntry.cs
```diff
@@ -0,0 +1,110 @@
...
+        /// <summary>
+        /// Returns a string that represents the current object.
+        /// </summary>
+        /// <returns>A string that represents the current object.</returns>
+        public override string ToString()
+        {
+            string contextFormatted = $"{{{String.Join("; ", Context.Select(item => $"[{item.Key}, {item.Value}]"))}}}";
+            return $"{DependencyType} {DependencyName} {DependencyData} named {TargetName} in {Duration} at {StartTime} " +
+                   $"(Successful: {IsSuccessful} - ResultCode: {ResultCode} - Context: {contextFormatted})";
+        }
```
fgheysels, line 108: What are we doing with these 'ToStrings' ?

fgheysels (commented):
Thx for picking this up so quickly.
I'm really excited to test this!

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
stijnmoreels: I don't the exact reason, but we have chosen to not use this one in our custom dependency types. It's mostly used by the built-in ones.

src/Arcus.Observability.Telemetry.Core/Logging/DependencyLogEntry.cs
stijnmoreels: Wel... the thing is... We can choose to use this Serilog formatting language, but that doesn't mean that every place it's used will be within Serilog. (That's not something we depent on or should be required when using our telemetry core package). So, if we don't wat our messages be: 
```csharp
Dependency Arcus.Observability.Telemetry.Core.Logging.DependencyLogTraget
```
we should have some way to represent it still as our 'original' sentence-like format, I guess.

stijnmoreels:
> Thx for picking this up so quickly.
> I'm really excited to test this!

You already can do this by using the MyGet package from this PR!

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
tomkerkhove: This should be added, see https://github.com/arcus-azure/arcus.observability/issues/165

fgheysels (approved):
LGTM I guess :)

fgheysels:
I'm going to use this myget version in one of my projects and see what it gives :)

src/Arcus.Observability.Telemetry.Core/Logging/DependencyLogEntry.cs
tomkerkhove: We are using our own type, but is the outcome the same if we:
1. Don't use a custom type
2. Don't override tostring

Because ideally we just allow people to use the standard approach imo

stijnmoreels: It's not entirely the same as the log properties will probably not be available as separate parts, but the rendered message will be the same.

fgheysels:
I'd like to test this, but it is impossible to use the preview built in my project(s).  (See discussion #211 ).
Are there other ways to be able to test this ?  (Maybe private build that is created by the regular pipeline ? )

tomkerkhove:
Let's get this PR done and we can create a pre-release for you to test with, ok for you?

docs/preview/features/writing-different-telemetry-types.md
tomkerkhove: Can you update the docs please?

src/Arcus.Observability.Telemetry.Core/Logging/DependencyLogEntry.cs
tomkerkhove: Ok, no problem

fgheysels:
> Let's get this PR done and we can create a pre-release for you to test with, ok for you?

Fine for me.

docs/preview/features/writing-different-telemetry-types.md
stijnmoreels: Done!

tomkerkhove (approved):
I trust @fgheysels's judgement

fgheysels:
Are we waiting for a test of the preview version for this ?
@stijnmoreels  @tomkerkhove  ?

PR #214, opened 2021-05-07
fix: remove telemetry prefix

stijnmoreels:
Removes the prefix when loging telemetry. This change requires us to look for other identifying properties when sinking the telemetry log event towards Azure Application Insights.

Relates to #205


src/Arcus.Observability.Telemetry.Core/MessageFormats.cs
```diff
@@ -11,7 +11,7 @@ public static class MessageFormats
         /// <summary>
         /// Gets the message format to log external dependencies; compatible with Application Insights 'Dependencies'.
         /// </summary>
-        public const string DependencyFormat = MessagePrefixes.Dependency + " {@" + ContextProperties.DependencyTracking.DependencyLogEntry + "}";
+        public const string DependencyFormat = "{@" + ContextProperties.DependencyTracking.DependencyLogEntry + "}";
```
fgheysels, line 14: Can't we get rid of these message-formats, and work with types as explained here 
https://github.com/arcus-azure/arcus.observability/issues/205#issuecomment-833453920

stijnmoreels: Not sure what you mean. We need these constants to retrieve their property value later on during the sinking to Application Insights and the telemetry type filtering.

docs/preview/features/writing-different-telemetry-types.md
```diff
@@ -64,7 +64,7 @@ durationMeasurement.Start();
 var startTime = DateTimeOffset.UtcNow;
```
tomkerkhove, line 65: I guess we'll have to update this once the other PR is in that adds the type of telemetry to the context

src/Arcus.Observability.Telemetry.Core/MessageFormats.cs
tomkerkhove: For telemetry type filtering I don't believe we need this since we have the new context property.

For the sinking, I think we could just re-use that and get the main information back based on the name, no?

src/Arcus.Observability.Telemetry.Serilog.Sinks.ApplicationInsights/Converters/ApplicationInsightsTelemetryConverter.cs
```diff
@@ -48,32 +40,22 @@ public override IEnumerable<ITelemetry> Convert(LogEvent logEvent, IFormatProvid
                 return _exceptionTelemetryConverter.Convert(logEvent, formatProvider);
             }
 
-            if (logEvent.MessageTemplate.Text.StartsWith(MessagePrefixes.RequestViaHttp))
+            if (logEvent.Properties.ContainsKey(ContextProperties.RequestTracking.RequestLogEntry))
```
tomkerkhove, line 43: We should rely on the new approach: https://github.com/arcus-azure/arcus.observability/pull/215/files#diff-0defdc0c79223504c82f4ce48ee49faccfb8fdf16640651e1747420d5ded944eR52

src/Arcus.Observability.Telemetry.Core/MessageFormats.cs
tomkerkhove: See https://github.com/arcus-azure/arcus.observability/pull/215/files#diff-0defdc0c79223504c82f4ce48ee49faccfb8fdf16640651e1747420d5ded944eR52

fgheysels: When you would log as is explained [here](https://github.com/arcus-azure/arcus.observability/issues/205#issuecomment-833453920), you don't need that 'message format' anymore.   In the AI sink, you'd be able to deserialize the string to a type, and retrieve all the properties from that type.

src/Arcus.Observability.Telemetry.Serilog.Sinks.ApplicationInsights/Converters/ApplicationInsightsTelemetryConverter.cs
stijnmoreels: But the context is part of the logentry type.. so we do have to use the message format to retrieve this entry, right? 
And since we already know by the kind of log entry that has been written, why would we have to deserialize the log entry into a Serology value, retrieve the context telemetry from that value, retrieve the telemetry type value from that context to get something we already know?

fgheysels: I'm not following anymore.  I think a short offline discussion can remove the clouds :)

fgheysels:
Are there still any pending / blocking issues  ?

stijnmoreels:
> Are there still any pending / blocking issues ?

Well, there's still some misunderstanding on the 'message formats' that's used for the telemetry logging; and using these formats to later determine in the sink which telemetry was logged.

fgheysels:
> > Are there still any pending / blocking issues ?
> 
> Well, there's still some misunderstanding on the 'message formats' that's used for the telemetry logging; and using these formats to later determine in the sink which telemetry was logged.

This has been cleared out in an offline meeting.  (Or is it online, when it's a Teams meeting :) )

src/Arcus.Observability.Telemetry.Core/MessageFormats.cs
fgheysels: OK, now I understand.  This string is used for the 'serialization' type of logging in Serilog:

```csharp
_logger.LogWarning(MetricFormat, metricVariable);
```

fgheysels (approved):

tomkerkhove (approved):

tomkerkhove:
@stijnmoreels Feel free to merge if everything has been resolved, looks good to us.

fgheysels (approved):

PR #215, opened 2021-05-07
feat: revisit telemetry type filter usage outside Azure Application Insights Serilog sink

### LINKED ISSUE #206: Provide support for filtering based on telemetry type with all sinks
Today, it seems like you can only filter for Azure Application Insights while I would like to do this for console logging.

It might be just me though so just checking but based on the code I think it doesn't work.

Provide support for filtering based on telemetry type with all sinks, instead of just Azure Application Insights.

image

**Describe alternatives you've considered**

**Additional context**

stijnmoreels:
Updates the `TelemetryTypeFilter` so it can be used with the new telemetry log format + outside Azure Application Insights sinks.

Closes #206

### DIFF
--- a/src/Arcus.Observability.Telemetry.Core/Logging/DependencyLogEntry.cs
+++ b/src/Arcus.Observability.Telemetry.Core/Logging/DependencyLogEntry.cs
@@ -49,6 +49,7 @@ namespace Arcus.Observability.Telemetry.Core.Logging
             StartTime = startTime.ToString(FormatSpecifiers.InvariantTimestampFormat);
             Duration = duration;
             Context = context;
+            Context[ContextProperties.General.TelemetryType] = TelemetryType.Dependency;
         }
 
         /// <summary>
--- a/src/Arcus.Observability.Telemetry.Core/TelemetryType.cs
+++ b/src/Arcus.Observability.Telemetry.Core/TelemetryType.cs
@@ -29,7 +29,7 @@ namespace Arcus.Observability.Telemetry.Core
         Events = 4,
         
         /// <summary>
-        /// Specifies the type of logged telemetry as value metrics.
+        /// Specifies the type of logged telemetry as metrics.
         /// </summary>
         Metrics = 8
     }
--- a/src/Arcus.Observability.Telemetry.Core/TelemetryType.cs
+++ b/src/Arcus.Observability.Telemetry.Core/TelemetryType.cs
@@ -1,14 +1,36 @@
-﻿namespace Arcus.Observability.Telemetry.Core
+﻿using Microsoft.Extensions.Logging;
+
+namespace Arcus.Observability.Telemetry.Core
 {
     /// <summary>
-    ///     Different types of telemetry
+    /// Represents the supported types of telemetry available to track during extensions on the <see cref="ILogger"/>.
+    /// Compatible with Azure Application Insights.
     /// </summary>
     public enum TelemetryType
     {
-        Trace,
-        Dependency,
-        Request,
-        Events,
-        Metrics
+        /// <summary>
+        /// Specifies the type of logged telemetry as traces.
+        /// </summary>
+        Trace = 0,
+        
+        /// <summary>
+        /// Specifies the type of logged telemetry as tracked dependencies.
+        /// </summary>
+        Dependency = 1,
+        
+        /// <summary>
+        /// Specifies the type of logged telemetry as HTTP requests.
+        /// </summary>
+        Request = 2,
+        
+        /// <summary>
+        /// Specifies the type of logged telemetry as custom events.
+        /// </summary>
+        Events = 4,
+        
+        /// <summary>
+        /// Specifies the type of logged telemetry as value metrics.
+        /// </summary>
+        Metrics = 8
     }
 }
--- a/src/Arcus.Observability.Telemetry.Core/TelemetryType.cs
+++ b/src/Arcus.Observability.Telemetry.Core/TelemetryType.cs
@@ -3,8 +3,8 @@
 namespace Arcus.Observability.Telemetry.Core
 {
     /// <summary>
-    /// Represents the supported types of telemetry available to track during extensions on the <see cref="ILogger"/>.
-    /// Compatible with Azure Application Insights.
+    /// <para>Represents the supported types of telemetry available to track during extensions on the <see cref="ILogger"/>.</para>
+    /// <para>Also compatible (but not limited to) with Azure Application Insights (see: <see href="https://observability.arcus-azure.net/features/sinks/azure-application-insights" /> for more information).</para>
     /// </summary>
     public enum TelemetryType
     {
--- a/src/Arcus.Observability.Telemetry.Serilog.Filters/TelemetryTypeFilter.cs
+++ b/src/Arcus.Observability.Telemetry.Serilog.Filters/TelemetryTypeFilter.cs
@@ -1,4 +1,6 @@
-﻿using Arcus.Observability.Telemetry.Core;
+﻿using System;
+using Arcus.Observability.Telemetry.Core;
+using GuardNet;
 using Serilog.Core;
 using Serilog.Events;
 
@@ -9,13 +11,11 @@ namespace Arcus.Observability.Telemetry.Serilog.Filters
     /// </summary>
     public class TelemetryTypeFilter : ILogEventFilter
     {
-        /// <summary>
-        ///     Constructor
-        /// </summary>
-        /// <param name="telemetryType">Current telemetry type that should be filtered</param>
-        /// <param name="isTrackingEnabled">Indication whether or not this telemetry type should be tracked</param>
         private TelemetryTypeFilter(TelemetryType telemetryType, bool? isTrackingEnabled)
         {
+            Guard.For(() => !Enum.IsDefined(typeof(TelemetryType), telemetryType), 
+                new ArgumentOutOfRangeException(nameof(telemetryType), telemetryType, "Requires a type of telemetry that's within the supported value range of the enumeration"));
+            
             TelemetryType = telemetryType;
             IsTrackingEnabled = isTrackingEnabled;
         }
@@ -34,8 +34,12 @@ namespace Arcus.Observability.Telemetry.Serilog.Filters
         ///     Provides a telemetry filter based on its type
         /// </summary>
         /// <param name="telemetryType">Current telemetry type that should be filtered</param>
+        /// <exception cref="ArgumentOutOfRangeException">Thrown when the <paramref name="telemetryType"/> is outside the bounds of the enumeration.</exception>
         public static TelemetryTypeFilter On(TelemetryType telemetryType)
         {
+            Guard.For(() => !Enum.IsDefined(typeof(TelemetryType), telemetryType), 
+                new ArgumentOutOfRangeException(nameof(telemetryType), telemetryType, "Requires a type of telemetry that's within the supported value range of the enumeration"));
+            
             return new TelemetryTypeFilter(telemetryType, isTrackingEnabled: null);
         }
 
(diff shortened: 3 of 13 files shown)


src/Arcus.Observability.Telemetry.Core/Logging/DependencyLogEntry.cs
```diff
@@ -49,6 +49,7 @@ public class DependencyLogEntry
             StartTime = startTime.ToString(FormatSpecifiers.InvariantTimestampFormat);
             Duration = duration;
             Context = context;
+            Context[ContextProperties.General.TelemetryType] = TelemetryType.Dependency;
```
stijnmoreels, line 52: Added this, not for functionality, but because it seemed something good to have, since when we format to JSON, there's otherwise no single property that can be used to determine which type it is.
But again, it's not being used in our functionality. I'll remove this if you think this is unnecessary.

src/Arcus.Observability.Telemetry.Serilog.Filters/TelemetryTypeFilter.cs
```diff
@@ -44,27 +48,36 @@ public static TelemetryTypeFilter On(TelemetryType telemetryType)
...
+                case TelemetryType.Dependency:
+                    return logEvent.Properties.ContainsKey(ContextProperties.DependencyTracking.DependencyLogEntry);
+                case TelemetryType.Request:
+                    return logEvent.Properties.ContainsKey(ContextProperties.RequestTracking.RequestLogEntry);
+                case TelemetryType.Events:
+                    return logEvent.Properties.ContainsKey(ContextProperties.EventTracking.EventLogEntry);
+                case TelemetryType.Metrics:
+                    return logEvent.Properties.ContainsKey(ContextProperties.MetricTracking.MetricLogEntry);
+                default:
+                    return false;
```
stijnmoreels, line 79: Problem here: the `TelemetryType` also includes `Trace`, but since we don't do traces ourselves, everything outside these cases is considered being a trace. Not sure if I should remove the enum value or add a case here and always return `false`.

src/Arcus.Observability.Tests.Integration/Serilog/TelemetryTypeFilterTests.cs
```diff
@@ -0,0 +1,1266 @@
...
+using ILogger = Microsoft.Extensions.Logging.ILogger;
+
+namespace Arcus.Observability.Tests.Integration.Serilog
+{
+    [Trait("Category", "Integration")]
+    public class TelemetryTypeFilterTests
+    {
+        private readonly Faker _bogusGenerator = new Faker();
+
+        public static IEnumerable<object[]> TelemetryTypesWithoutEvent => GetTelemetryTypesWithout(TelemetryType.Events);
```
stijnmoreels, line 28: Oh, it really itched to do some property-based testing... 😄

stijnmoreels:
I'll add @fgheysels from now on too on each PR, as he's promoted to CodeOwner?

src/Arcus.Observability.Telemetry.Core/Logging/DependencyLogEntry.cs
fgheysels: I currently have no opinion on that. :)  If I see the result, I can maybe form an opinion :)

src/Arcus.Observability.Telemetry.Serilog.Filters/TelemetryTypeFilter.cs
```diff
@@ -44,27 +48,36 @@ public static TelemetryTypeFilter On(TelemetryType telemetryType)
...
+                new ArgumentOutOfRangeException(nameof(telemetryType), telemetryType, "Requires a type of telemetry that's within the supported value range of the enumeration"));
+            
             return new TelemetryTypeFilter(telemetryType, isTrackingEnabled);
         }
 
         /// <inheritdoc />
         public bool IsEnabled(LogEvent logEvent)
         {
-            if (IsTrackingEnabled == true)
+            if (IsTrackingEnabled is true)
```
fgheysels, line 63: What's the advantage of writing `is true` instead of `== true` ?

src/Arcus.Observability.Telemetry.Serilog.Filters/TelemetryTypeFilter.cs
fgheysels: When is this method called ?  I would assume that it's called to determine if the event must be logged or not ?  Isnt that done based on the configured loglevel ?  Or am I talking rubbish here ?

fgheysels (commented):
Is the way events / metrics / etc.... are written now being changed, since I haven't seen any changes on that, or did I miss something ?

src/Arcus.Observability.Telemetry.Serilog.Filters/TelemetryTypeFilter.cs
stijnmoreels: `is` forces the terms to be constants l, that's also why `... is null` is much safer than `... == null` bc the `==` can be overwritten. It's a personal choice maybe or my paranoia speaking. I tend to use `is` where possible.

src/Arcus.Observability.Telemetry.Core/Logging/DependencyLogEntry.cs
tomkerkhove: This is fine for me, as long as the Application Insights sink removes this for that telemetry, other sinks can have this without a problem imo

src/Arcus.Observability.Telemetry.Core/TelemetryType.cs
```diff
@@ -1,14 +1,36 @@
...
+        /// </summary>
+        Request = 2,
+        
+        /// <summary>
+        /// Specifies the type of logged telemetry as custom events.
+        /// </summary>
+        Events = 4,
+        
+        /// <summary>
+        /// Specifies the type of logged telemetry as value metrics.
```
tomkerkhove, line 32:
```suggestion
        /// Specifies the type of logged telemetry as metrics.
```

src/Arcus.Observability.Telemetry.Core/TelemetryType.cs
```diff
@@ -1,14 +1,36 @@
-﻿namespace Arcus.Observability.Telemetry.Core
+﻿using Microsoft.Extensions.Logging;
+
+namespace Arcus.Observability.Telemetry.Core
 {
     /// <summary>
-    ///     Different types of telemetry
+    /// Represents the supported types of telemetry available to track during extensions on the <see cref="ILogger"/>.
+    /// Compatible with Azure Application Insights.
```
tomkerkhove, line 7: It's compatible with more than Azure Application Insights, no? Maybe rephrase it so that we make it easier to use and add a link to our sink doc?

src/Arcus.Observability.Tests.Unit/Telemetry/TelemetryTypeFilterTests.cs
```diff
@@ -0,0 +1,114 @@
...
+        [InlineData(TelemetryType.Metrics | TelemetryType.Request | TelemetryType.Events)]
+        public void CreateFilter_WithTelemetryTypeOutsideEnum_Fails(TelemetryType telemetryType)
+        {
+            Assert.ThrowsAny<ArgumentException>(() => TelemetryTypeFilter.On(telemetryType));
+        }
+        
+        [Theory]
+        [InlineData(TelemetryType.Request | TelemetryType.Dependency)]
+        [InlineData(TelemetryType.Metrics | TelemetryType.Request | TelemetryType.Events)]
+        public void CreateFilterWithTracked_WithTelemetryTypeOutsideEnum_Fails(TelemetryType telemetryType)
```
tomkerkhove, line 105: I'm not sure if I get the test case here, can you elaborate please?

src/Arcus.Observability.Tests.Unit/Telemetry/TelemetryTypeFilterTests.cs
```diff
@@ -0,0 +1,114 @@
...
+            
+            // Act
+            bool isEnabled = filter.IsEnabled(logEvent);
+            
+            // Assert
+            Assert.False(isEnabled);
+        }
+
+        [Fact]
+        public void LogEventWithoutTelemetry_DoesntFilterAnything_Succeeds()
```
tomkerkhove, line 61: How is this test different from LogEventWithoutValidTelemetry_DoesntFilterAnything_Succeeds?

src/Arcus.Observability.Tests.Unit/Telemetry/TelemetryTypeFilterTests.cs
```diff
@@ -0,0 +1,114 @@
...
+            
+            // Act
+            bool isEnabled = filter.IsEnabled(logEvent);
+            
+            // Assert
+            Assert.True(isEnabled);
+        }
+
+        [Fact]
+        public void LogEventAsTelemetry_FiltersOutCorrectTelemetry_Succeeds()
```
tomkerkhove, line 39: I would add a test case like this one, but instead:
1. Tell it to filter a given telemetry type, for example dependency
2. Write a new log event with a different type
3. Entry is not filtered

src/Arcus.Observability.Telemetry.Serilog.Filters/TelemetryTypeFilter.cs
```diff
@@ -44,27 +48,36 @@ public static TelemetryTypeFilter On(TelemetryType telemetryType)
...
             {
                 return true;
             }
 
-            var telemetryType = logEvent.Properties.GetAsEnum<TelemetryType>(ContextProperties.General.TelemetryType);
-            if (telemetryType == null)
+            switch (TelemetryType)
```
tomkerkhove, line 68: Can't we revert this change now that we have this?

`ContextProperties.General.TelemetryType`?

stijnmoreels: No because the telemetry type is a property of the log entry type, not the Serilog log event.

src/Arcus.Observability.Tests.Unit/Telemetry/TelemetryTypeFilterTests.cs
stijnmoreels: Woops, yeah, was normally for invalid telemetry contexts but since we can use a more simpler approach, we can remove this one.

src/Arcus.Observability.Tests.Unit/Telemetry/TelemetryTypeFilterTests.cs
stijnmoreels: Yes, missed these ones. I'll add some theory for that. Thanks. Resolving.

src/Arcus.Observability.Tests.Unit/Telemetry/TelemetryTypeFilterTests.cs
stijnmoreels: The telemetry filter guards against misused telemetry type enumeration values. All the inputs must be within the bounds of the enumeration.
This test checks that values outside the scope of the enumeration will result in argument exceptions.

src/Arcus.Observability.Telemetry.Core/TelemetryType.cs
stijnmoreels: Hmm, yeah, could be used outside AI.

stijnmoreels: Updated this. Thanks!

src/Arcus.Observability.Telemetry.Serilog.Filters/TelemetryTypeFilter.cs
stijnmoreels: > When is this method called ?

Serilog filters are predicates on the log level. They're called before the the event is logged to the sink.

> I would assume that it's called to determine if the event must be logged or not ?

Yes.

>  Isnt that done based on the configured loglevel ?

Yes, also, but a filter can be added for extra more grainer control over the events being logged.

> Or am I talking rubbish here ?

Not at all 😉 .

tomkerkhove (approved):

fgheysels (approved):

PR #229, opened 2021-09-02
feat: Provide dependency names for dependencies & requests

pim-simons:
- Requests are now logged with an operation name so they don't show in Application Insights as '<Empty>'. For example: 
image
After discussion with @stijnmoreels we decided to add 2 overloads to be able to specify a custom operataion name as well. 
- Names have been added to all dependancies so that they don't show up in Application Insights 'n/a'. For example:
image

### DIFF
--- a/src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
+++ b/src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
@@ -141,7 +141,7 @@ namespace Microsoft.Extensions.Logging
             string resourcePath = request.RequestUri.AbsolutePath;
             string host = $"{request.RequestUri.Scheme}://{request.RequestUri.Host}";
 
-            logger.LogWarning(RequestFormat, new RequestLogEntry(request.Method.ToString(), host, resourcePath, null, statusCode, duration, context));
+            logger.LogWarning(RequestFormat, new RequestLogEntry(request.Method.ToString(), host, resourcePath, statusCode, duration, context));
         }
 
         /// <summary>
--- a/src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
+++ b/src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
@@ -225,6 +225,37 @@ namespace Microsoft.Extensions.Logging
         /// <param name="dependencyType">Custom type of dependency</param>
         /// <param name="dependencyData">Custom data of dependency</param>
         /// <param name="isSuccessful">Indication whether or not the operation was successful</param>
+        /// <param name="dependencyName">The name of the dependency</param>
+        /// <param name="measurement">Measuring the latency to call the dependency</param>
+        /// <param name="context">Context that provides more insights on the dependency that was measured</param>
+        /// <exception cref="ArgumentNullException">
+        ///     Thrown when the <paramref name="logger"/>, <paramref name="dependencyData"/>, <paramref name="measurement"/> is <c>null</c>.
+        /// </exception>
+        /// <exception cref="ArgumentException">Thrown when the <paramref name="dependencyData"/> is blank.</exception>
+        public static void LogDependency(
+            this ILogger logger,            
+            string dependencyType,
+            object dependencyData,
+            bool isSuccessful,
+            string dependencyName,
+            DependencyMeasurement measurement,
+            Dictionary<string, object> context = null)
+        {
+            Guard.NotNull(logger, nameof(logger), "Requires a logger instance to track telemetry");
+            Guard.NotNullOrWhitespace(dependencyType, nameof(dependencyType), "Requires a non-blank custom dependency type when tracking the custom dependency");
+            Guard.NotNull(dependencyData, nameof(dependencyData), "Requires custom dependency data when tracking the custom dependency");
+            Guard.NotNull(measurement, nameof(measurement), "Requires a dependency measurement instance to track the latency of the dependency when tracking the custom dependency");
+
+            LogDependency(logger, dependencyType, dependencyData, isSuccessful, dependencyName, measurement.StartTime, measurement.Elapsed, context);
+        }
+
+        /// <summary>
+        ///     Logs a dependency.
+        /// </summary>
+        /// <param name="logger">The logger to track the telemetry.</param>
+        /// <param name="dependencyType">Custom type of dependency</param>
+        /// <param name="dependencyData">Custom data of dependency</param>
+        /// <param name="isSuccessful">Indication whether or not the operation was successful</param>
         /// <param name="startTime">Point in time when the interaction with the dependency was started</param>
         /// <param name="duration">Duration of the operation</param>
         /// <param name="context">Context that provides more insights on the dependency that was measured</param>
@@ -248,6 +279,40 @@ namespace Microsoft.Extensions.Logging
             Guard.NotLessThan(duration, TimeSpan.Zero, nameof(duration), "Requires a positive time duration of the dependency operation");
             
             LogDependency(logger, dependencyType, dependencyData, targetName: null, isSuccessful: isSuccessful, startTime: startTime, duration: duration, context: context);
+        }
+
+        /// <summary>
+        ///     Logs a dependency.
+        /// </summary>
+        /// <param name="logger">The logger to track the telemetry.</param>
+        /// <param name="dependencyType">Custom type of dependency</param>
+        /// <param name="dependencyData">Custom data of dependency</param>
+        /// <param name="isSuccessful">Indication whether or not the operation was successful</param>
+        /// <param name="dependencyName">The name of the dependency</param>
+        /// <param name="startTime">Point in time when the interaction with the dependency was started</param>
+        /// <param name="duration">Duration of the operation</param>
+        /// <param name="context">Context that provides more insights on the dependency that was measured</param>
+        /// <exception cref="ArgumentNullException">
+        ///     Thrown when the <paramref name="logger"/>, <paramref name="dependencyData"/> is <c>null</c>.
+        /// </exception>
+        /// <exception cref="ArgumentException">Thrown when the <paramref name="dependencyData"/> is blank.</exception>
+        /// <exception cref="ArgumentOutOfRangeException">Thrown when the <paramref name="duration"/> is a negative time range.</exception>
+        public static void LogDependency(
+            this ILogger logger,            
+            string dependencyType,
+            object dependencyData,
+            bool isSuccessful,
+            string dependencyName,
+            DateTimeOffset startTime,
+            TimeSpan duration,
+            Dictionary<string, object> context = null)
+        {
+            Guard.NotNull(logger, nameof(logger), "Requires a logger instance to track telemetry");
+            Guard.NotNullOrWhitespace(dependencyType, nameof(dependencyType), "Requires a non-blank custom dependency type when tracking the custom dependency");
+            Guard.NotNull(dependencyData, nameof(dependencyData), "Requires custom dependency data when tracking the custom dependency");
+            Guard.NotLessThan(duration, TimeSpan.Zero, nameof(duration), "Requires a positive time duration of the dependency operation");
+
+            LogDependency(logger, dependencyType, dependencyData, targetName: null, isSuccessful: isSuccessful, dependencyName, startTime: startTime, duration: duration, context: context);
         }
 
         /// <summary>
@@ -287,6 +352,39 @@ namespace Microsoft.Extensions.Logging
         /// <param name="logger">The logger to track the telemetry.</param>
         /// <param name="dependencyType">Custom type of dependency</param>
         /// <param name="dependencyData">Custom data of dependency</param>
+        /// <param name="targetName">Name of the dependency target</param>
+        /// <param name="isSuccessful">Indication whether or not the operation was successful</param>
+        /// <param name="dependencyName">The name of the dependency</param>
+        /// <param name="measurement">Measuring the latency to call the dependency</param>
+        /// <param name="context">Context that provides more insights on the dependency that was measured</param>
+        /// <exception cref="ArgumentNullException">
+        ///     Thrown when the <paramref name="logger"/>, <paramref name="dependencyData"/>, <paramref name="measurement"/> is <c>null</c>.
+        /// </exception>
+        /// <exception cref="ArgumentException">Thrown when the <paramref name="dependencyData"/> is blank.</exception>
+        public static void LogDependency(
+            this ILogger logger,
+            string dependencyType,
+            object dependencyData,
+            string targetName,
+            bool isSuccessful,
+            string dependencyName,
+            DependencyMeasurement measurement,
+            Dictionary<string, object> context = null)
+        {
+            Guard.NotNull(logger, nameof(logger), "Requires a logger instance to track telemetry");
+            Guard.NotNullOrWhitespace(dependencyType, nameof(dependencyType), "Requires a non-blank custom dependency type when tracking the custom dependency");
+            Guard.NotNull(dependencyData, nameof(dependencyData), "Requires custom dependency data when tracking the custom dependency");
+            Guard.NotNull(measurement, nameof(measurement), "Requires a dependency measurement instance to track the latency of the dependency when tracking the custom dependency");
+
+            LogDependency(logger, dependencyType, dependencyData, targetName, isSuccessful, dependencyName, measurement.StartTime, measurement.Elapsed, context);
+        }
+
+        /// <summary>
+        ///     Logs a dependency.
+        /// </summary>
+        /// <param name="logger">The logger to track the telemetry.</param>
+        /// <param name="dependencyType">Custom type of dependency</param>
+        /// <param name="dependencyData">Custom data of dependency</param>
         /// <param name="targetName">Name of dependency target</param>
         /// <param name="isSuccessful">Indication whether or not the operation was successful</param>
         /// <param name="startTime">Point in time when the interaction with the dependency was started</param>
@@ -311,18 +409,54 @@ namespace Microsoft.Extensions.Logging
             Guard.NotNullOrWhitespace(dependencyType, nameof(dependencyType), "Requires a non-blank custom dependency type when tracking the custom dependency");
             Guard.NotNull(dependencyData, nameof(dependencyData), "Requires custom dependency data when tracking the custom dependency");
             Guard.NotLessThan(duration, TimeSpan.Zero, nameof(duration), "Requires a positive time duration of the dependency operation");
-            
+
+            LogDependency(logger, dependencyType, dependencyData, targetName, isSuccessful, dependencyName: targetName, startTime, duration, context);
+        }
+
+        /// <summary>
+        ///     Logs a dependency.
+        /// </summary>
+        /// <param name="logger">The logger to track the telemetry.</param>
+        /// <param name="dependencyType">Custom type of dependency</param>
+        /// <param name="dependencyData">Custom data of dependency</param>
+        /// <param name="targetName">Name of dependency target</param>
+        /// <param name="isSuccessful">Indication whether or not the operation was successful</param>
+        /// <param name="dependencyName">The name of the dependency</param>
+        /// <param name="startTime">Point in time when the interaction with the dependency was started</param>
+        /// <param name="duration">Duration of the operation</param>
+        /// <param name="context">Context that provides more insights on the dependency that was measured</param>
+        /// <exception cref="ArgumentNullException">
+        ///     Thrown when the <paramref name="logger"/>, <paramref name="dependencyData"/> is <c>null</c>.
+        /// </exception>
+        /// <exception cref="ArgumentException">Thrown when the <paramref name="dependencyData"/> is blank.</exception>
+        /// <exception cref="ArgumentOutOfRangeException">Thrown when the <paramref name="duration"/> is a negative time range.</exception>
+        public static void LogDependency(
+            this ILogger logger,            
+            string dependencyType,
+            object dependencyData,
+            string targetName,
+            bool isSuccessful,
+            string dependencyName,
+            DateTimeOffset startTime,
+            TimeSpan duration,
+            Dictionary<string, object> context = null)
+        {
+            Guard.NotNull(logger, nameof(logger), "Requires a logger instance to track telemetry");
+            Guard.NotNullOrWhitespace(dependencyType, nameof(dependencyType), "Requires a non-blank custom dependency type when tracking the custom dependency");
+            Guard.NotNull(dependencyData, nameof(dependencyData), "Requires custom dependency data when tracking the custom dependency");
+            Guard.NotLessThan(duration, TimeSpan.Zero, nameof(duration), "Requires a positive time duration of the dependency operation");
+
             context = context ?? new Dictionary<string, object>();
 
             logger.LogWarning(DependencyFormat, new DependencyLogEntry(
-                dependencyType, 
-                dependencyName: targetName, 
-                dependencyData: dependencyData, 
-                targetName: targetName, 
-                duration: duration, 
-                startTime: startTime, 
+                dependencyType,
+                dependencyName: dependencyName,
+                dependencyData: dependencyData,
+                targetName: targetName,
+                duration: duration,
+                startTime: startTime,
                 resultCode: null,
-                isSuccessful: isSuccessful, 
+                isSuccessful: isSuccessful,
                 context: context));
         }
 
--- a/src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
+++ b/src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
@@ -316,7 +316,7 @@ namespace Microsoft.Extensions.Logging
 
             logger.LogWarning(DependencyFormat, new DependencyLogEntry(
                 dependencyType, 
-                dependencyName: null, 
+                dependencyName: targetName, 
                 dependencyData: dependencyData, 
                 targetName: targetName, 
                 duration: duration, 
@@ -398,10 +398,10 @@ namespace Microsoft.Extensions.Logging
             Guard.NotLessThan(duration, TimeSpan.Zero, nameof(duration), "Requires a positive time duration of the Azure Key Vault operation");
             
             context = context ?? new Dictionary<string, object>();
-            
+
             logger.LogWarning(DependencyFormat, new DependencyLogEntry(
                 dependencyType: "Azure key vault", 
-                dependencyName: null, 
+                dependencyName: vaultUri, 
                 dependencyData: secretName, 
                 targetName: vaultUri, 
                 duration: duration, 
@@ -470,7 +470,7 @@ namespace Microsoft.Extensions.Logging
 
             logger.LogWarning(DependencyFormat, new DependencyLogEntry(
                 dependencyType: "Azure Search", 
-                dependencyName: null, 
+                dependencyName: searchServiceName, 
                 dependencyData: operationName, 
                 targetName: searchServiceName, 
                 duration: duration, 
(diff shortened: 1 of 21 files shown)


src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
```diff
@@ -231,7 +316,7 @@ public static class ILoggerExtensions
 
             logger.LogWarning(DependencyFormat, new DependencyLogEntry(
                 dependencyType, 
-                dependencyName: null, 
+                dependencyName: targetName,
```
pim-simons, line 319: This is one I am not sure about, not sure if targetName is the right value for the dependencyName. 
Maybe it is better to add an overload where dependencyName can be specified and otherwise fall back on targetName?

src/Arcus.Observability.Telemetry.Serilog.Sinks.ApplicationInsights/Converters/CloudContextConverter.cs
```diff
@@ -29,6 +30,16 @@ public void EnrichWithAppInfo(LogEvent logEvent, ITelemetry telemetry)
...
+
+            //if (telemetry is RequestTelemetry f)
+            //{
+            //    f.Context.Operation.Name = f.Name;
+            //}
+
+            //if (telemetry is DependencyTelemetry d)
+            //{
+            //    d.Context.Operation.Name = d.Name;
+            //}
```
tomkerkhove, line 42:
```suggestion
```

tomkerkhove (commented):
If the PR is only for the request, can we remove the changes for the dependencies please?

tomkerkhove:
If you keep the dependency changes, can you please use the convention from https://github.com/arcus-azure/arcus-service-to-service-correlation-poc/blob/d1526c963c04b35243d71a65f9eefd3eeba1dcee/src/Arcus.POC.Observability.Telemetry.Serilog.Sinks.ApplicationInsights/Extensions/ILoggerExtensions.cs#L118 for Service Bus?

I didn't include the test but the goal is to align with the official SDK notation.

pim-simons:
> 
> 
> If you keep the dependency changes, can you please use the convention from https://github.com/arcus-azure/arcus-service-to-service-correlation-poc/blob/d1526c963c04b35243d71a65f9eefd3eeba1dcee/src/Arcus.POC.Observability.Telemetry.Serilog.Sinks.ApplicationInsights/Extensions/ILoggerExtensions.cs#L118 for Service Bus?
> 
> I didn't include the test but the goal is to align with the official SDK notation.

Thanks for the feedback! 
Checked the code, the requested convention is `$"{serviceBusNamespaceEndpoint} | {queueName}"` but the Service Bus Namespace is currently not available on the Service Bus logging methods, only the type (Queue, Topic or Unknown) and name of the entity. 
I could create new overloads for the Service Bus logging to include the namespace but then how do we set the dependencyName for the current methods without the namespace? Maybe we could do something like `$"{entityType} | {entityName}"` instead of just the entityName? 
Let me know if I missed anything!

stijnmoreels:
I'll have a look later, as I'm not allowed to work long for Arcus for more than a day in a week 😅, (shame).

Thanks for your PR! Very much appreciated! 🙏🥇

tomkerkhove:
> > If you keep the dependency changes, can you please use the convention from [arcus-azure/arcus-service-to-service-correlation-poc@`d1526c9`/src/Arcus.POC.Observability.Telemetry.Serilog.Sinks.ApplicationInsights/Extensions/ILoggerExtensions.cs#L118](https://github.com/arcus-azure/arcus-service-to-service-correlation-poc/blob/d1526c963c04b35243d71a65f9eefd3eeba1dcee/src/Arcus.POC.Observability.Telemetry.Serilog.Sinks.ApplicationInsights/Extensions/ILoggerExtensions.cs#L118) for Service Bus?
> > I didn't include the test but the goal is to align with the official SDK notation.
> 
> Thanks for the feedback!
> Checked the code, the requested convention is `$"{serviceBusNamespaceEndpoint} | {queueName}"` but the Service Bus Namespace is currently not available on the Service Bus logging methods, only the type (Queue, Topic or Unknown) and name of the entity.
> I could create new overloads for the Service Bus logging to include the namespace but then how do we set the dependencyName for the current methods without the namespace? Maybe we could do something like `$"{entityType} | {entityName}"` instead of just the entityName?
> Let me know if I missed anything!

We should have them indeed, but let's use a separate PR for that.

src/Arcus.Observability.Telemetry.Serilog.Sinks.ApplicationInsights/Converters/RequestTelemetryConverter.cs
```diff
@@ -45,6 +46,15 @@ protected override RequestTelemetry CreateTelemetryEntry(LogEvent logEvent, IFor
...
             };
 
+            if (!String.IsNullOrEmpty(operationName))
+            {
+                requestTelemetry.Context.Operation.Name = $"{requestMethod} {operationName}";
+            }
+            else
+            {
+                requestTelemetry.Context.Operation.Name = requestName;
+            }
```
tomkerkhove, line 56: Since we need this for dependencies as well I think we should move this to `OperationContextConverter` similar to this approach:
https://github.com/arcus-azure/arcus-service-to-service-correlation-poc/blob/d1526c963c04b35243d71a65f9eefd3eeba1dcee/src/Arcus.POC.Observability.Telemetry.Serilog.Sinks.ApplicationInsights/Converters/OperationContextConverter.cs#L34-L59

There you can do the same logic for the request similar to what you do here

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
tomkerkhove: Agreed, then this one can call the new one and pass the targetName indeed but I'd consider deprecating this extension in favor of the new one.

tomkerkhove (changes requested):
Looks like a good start, thank you!

I see two things we should do:
- Provide unit tests for dependencies
- Move the conversion to our `OperationContextConverter`

src/Arcus.Observability.Telemetry.Serilog.Sinks.ApplicationInsights/Converters/RequestTelemetryConverter.cs
pim-simons: Seems like a better place to put it indeed. Will take a look and make the changes.
In your example it is placed within `EnrichWithCorrelationInfo` but as far as I can check the operation name doesn't really  have anything to do with correlation (at least I can't find it in https://docs.microsoft.com/en-us/azure/azure-monitor/app/correlation). Might be better to put it in a new method like `EnrichWithOperationName`?

tomkerkhove: This is just the method that is being called, feel free to create a new one indeed.

Thanks!

src/Arcus.Observability.Telemetry.Core/Logging/RequestLogEntry.cs
```diff
@@ -32,6 +33,7 @@ public class RequestLogEntry
             string method, 
             string host, 
             string uri, 
+            string operationName,
```
stijnmoreels, line 36: Technically, this is a breaking change, as it's a `public` class with a `public` ctor. We should consider making an overload instead.

src/Arcus.Observability.Telemetry.Serilog.Sinks.ApplicationInsights/Converters/RequestTelemetryConverter.cs
stijnmoreels: Ah, yes, we could do it there too. Requires some casting though, while here it does not. Same for me! We should be careful of scattering these things around instead of placing combining them in the same type.

src/Arcus.Observability.Tests.Integration/Serilog/Sinks/ApplicationInsights/RequestTests.cs
```diff
@@ -23,6 +23,81 @@ public RequestTests(ITestOutputHelper outputWriter) : base(outputWriter)
...
+                    Assert.Contains(results.Value, result => result.Request.Url == $"{requestUri.Scheme}://{requestUri.Host}{requestUri.AbsolutePath}");
+                    Assert.Contains(results.Value, result => !String.IsNullOrEmpty(result.Operation.Name));
+                });
+            }
+
+            VerifyLogEventProperties(requestUri);
+        }
+
+        [Fact]
+        public async Task LogRequest_SinksToApplicationInsightsWithOperationName_ResultsInRequestTelemetry()
```
stijnmoreels, line 64: Great stuff! Shows how we can pass along an operation name, now. Since the other dependency logging extensions have been altered to include a dependency name, we should also test for those. (unless you want to split the other dependencies in another PR).
Normally, we have for every logging extension an integration test, so in some cases you could only add an extra assertion for the dependency name, without having to write a dedicated test like these ones.
Thanks again!

stijnmoreels (changes requested):
Great first step! You really have successfully found the necessary places to incorporate your requested changes. Like Tom said, we usually provide unit tests for every publicly available member (I'm sure you'll find the logging unit tests for requests).

- [x] Provide unit tests for requests (and dependencies if you don't create a separate PR for them)
- [ ] Update the dependency integration tests to include the passed-along operation name

Let us know @pim-simons , if you need any help! You're doing a great job!

src/Arcus.Observability.Telemetry.Core/Logging/RequestLogEntry.cs
pim-simons: I added an overload, see https://github.com/arcus-azure/arcus.observability/pull/229/commits/13c77ee974d22ae0d69215b57ad32405118f33e5

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
pim-simons: Added an overload so we can specify the dependencyName on the LogDependency, added some tests as well, see https://github.com/arcus-azure/arcus.observability/pull/229/commits/058865f2848428ff361e7ba2c033bae9becd83e1

src/Arcus.Observability.Telemetry.Serilog.Sinks.ApplicationInsights/Converters/RequestTelemetryConverter.cs
pim-simons: Moved it to a new method in the `OperationContextConverter` called `EnrichWithOperationName`, see https://github.com/arcus-azure/arcus.observability/pull/229/commits/66ceec04cc15457cba3d119c42739df437c9b1ff

Not sure if the implementation is completely correct as I still set the OperationName in the `RequestTelemetryConverter` if it has been specified with a custom value. For the RequestTelemetry I then check in the `OperationContextConverter` if the OperationName has already been set and if not fallback to the name of the telemetryEntry. 

If this is not according to Arcus guidelines or standards I could use some tips on how to implement correctly 🙏🏻

src/Arcus.Observability.Tests.Integration/Serilog/Sinks/ApplicationInsights/RequestTests.cs
pim-simons: I added the dependency tests to check for the dependency name. Added some tests as well, let me know if tests are still missing 👍🏻

src/Arcus.Observability.Tests.Integration/Serilog/Sinks/ApplicationInsights/CustomDependencyTests.cs
```diff
@@ -61,5 +61,53 @@ public async Task LogDependency_SinksToApplicationInsights_ResultsInDependencyTe
...
+                {
+                    EventsResults<EventsDependencyResult> results = await client.Events.GetDependencyEventsAsync(ApplicationId, timespan: "PT30M");
+                    Assert.NotEmpty(results.Value);
+                    Assert.Contains(results.Value, result => result.Dependency.Type == dependencyType && result.Dependency.Data == dependencyData && result.Dependency.Name == dependencyName);
+                });
+            }
+
+            AssertX.Any(GetLogEventsFromMemory(), logEvent => {
+                StructureValue logEntry = logEvent.Properties.GetAsStructureValue(ContextProperties.DependencyTracking.DependencyLogEntry);
+                Assert.NotNull(logEntry);
```
stijnmoreels, line 98: Nit
```suggestion
AssertX.Any(GetLogEventsFromMemory(), logEvent => 
{
                StructureValue logEntry = logEvent.Properties.GetAsStructureValue(ContextProperties.DependencyTracking.DependencyLogEntry);
                Assert.NotNull(logEntry);
```

src/Arcus.Observability.Tests.Unit/Telemetry/ILoggerExtensionsTests.cs
```diff
@@ -914,6 +1050,9 @@ public void LogBlobStorageDependencyWithDependencyMeasurement_ValidArguments_Suc
             Assert.Contains(isSuccessful.ToString(), logMessage);
             Assert.Contains(startTime.ToString(FormatSpecifiers.InvariantTimestampFormat), logMessage);
             Assert.Contains(duration.ToString(), logMessage);
+            // Check specifically on dependencyName
+            string dependencyName = $"{accountName}/{containerName}";
+            Assert.Contains("Azure blob " + dependencyName, logMessage);
```
stijnmoreels, line 1055: If we move the dependency name to a dedicated variable, there's no need to specify it via a comment, right?
```suggestion
            string dependencyName = $"{accountName}/{containerName}";
            Assert.Contains("Azure blob " + dependencyName, logMessage);
```

src/Arcus.Observability.Tests.Unit/Telemetry/ILoggerExtensionsTests.cs
```diff
@@ -803,6 +931,8 @@ public void LogServiceBusTopicDependencyWithDependencyMeasurement_ValidArguments
             Assert.Contains(startTime.ToString(FormatSpecifiers.InvariantTimestampFormat), logMessage);
             Assert.Contains(duration.ToString(), logMessage);
             Assert.Contains(isSuccessful.ToString(), logMessage);
+            // Check specifically on dependencyName
+            Assert.Contains("Azure Service Bus " + topicName, logMessage);
```
stijnmoreels, line 935: Not a big fan of comments like this 😄 . Maybe if we move it to a dedicated variable, there's no need?
```suggestion
            Assert.Contains(isSuccessful.ToString(), logMessage);
            var dependencyName = topicName;
            Assert.Contains("Azure Service Bus " + dependencyName, logMessage);
```

stijnmoreels (commented):
We're almost there! Let's have @tomkerkhove another look, especially on the casting/child classes. Otherwise, great job! 🥇

src/Arcus.Observability.Tests.Integration/Serilog/Sinks/ApplicationInsights/RequestTests.cs
stijnmoreels: I think there' all there. Both unit and integration tests updated. Great!

src/Arcus.Observability.Tests.Unit/Telemetry/ILoggerExtensionsTests.cs
pim-simons: No problem 👍🏻  move the comments to dedicated variables, see: https://github.com/arcus-azure/arcus.observability/pull/229/commits/e821504de70e61733938d1a75efcee6498505887

stijnmoreels (approved):
I'll approve on my part. Great stuff!

src/Arcus.Observability.Telemetry.Core/Logging/RequestLogEntry.cs
```diff
@@ -28,10 +28,53 @@ public class RequestLogEntry
         ///     Thrown when the <paramref name="statusCode"/>'s status code is outside the 100-599 inclusively,
         ///     the <paramref name="duration"/> is a negative time range.
         /// </exception>
+        public RequestLogEntry(
```
tomkerkhove, line 31: Can't we call our other constructor to reduce duplication?

tomkerkhove: We should maybe assign a default operation name here when calling the other constructor

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
```diff
@@ -133,6 +218,37 @@ public static class ILoggerExtensions
...
+        ///     Thrown when the <paramref name="logger"/>, <paramref name="dependencyData"/>, <paramref name="measurement"/> is <c>null</c>.
+        /// </exception>
+        /// <exception cref="ArgumentException">Thrown when the <paramref name="dependencyData"/> is blank.</exception>
+        public static void LogDependency(
+            this ILogger logger,            
+            string dependencyType,
+            object dependencyData,
+            bool isSuccessful,
+            string dependencyName,
+            DependencyMeasurement measurement,
```
tomkerkhove, line 241: This feels more natural to me 
```suggestion
            string dependencyType,
            string dependencyName,
            object dependencyData,
            bool isSuccessful,
            DependencyMeasurement measurement,
```

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
```diff
@@ -165,6 +281,40 @@ public static class ILoggerExtensions
...
+        ///     Thrown when the <paramref name="logger"/>, <paramref name="dependencyData"/> is <c>null</c>.
+        /// </exception>
+        /// <exception cref="ArgumentException">Thrown when the <paramref name="dependencyData"/> is blank.</exception>
+        /// <exception cref="ArgumentOutOfRangeException">Thrown when the <paramref name="duration"/> is a negative time range.</exception>
+        public static void LogDependency(
+            this ILogger logger,            
+            string dependencyType,
+            object dependencyData,
+            bool isSuccessful,
+            string dependencyName,
```
tomkerkhove, line 305: Same here, I'd change the ordering

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
```diff
@@ -196,6 +346,39 @@ public static class ILoggerExtensions
...
+        ///     Thrown when the <paramref name="logger"/>, <paramref name="dependencyData"/>, <paramref name="measurement"/> is <c>null</c>.
+        /// </exception>
+        /// <exception cref="ArgumentException">Thrown when the <paramref name="dependencyData"/> is blank.</exception>
+        public static void LogDependency(
+            this ILogger logger,
+            string dependencyType,
+            object dependencyData,
+            string targetName,
+            bool isSuccessful,
+            string dependencyName,
```
tomkerkhove, line 370: Ordering

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
```diff
@@ -226,18 +409,54 @@ public static class ILoggerExtensions
...
+        /// </exception>
+        /// <exception cref="ArgumentException">Thrown when the <paramref name="dependencyData"/> is blank.</exception>
+        /// <exception cref="ArgumentOutOfRangeException">Thrown when the <paramref name="duration"/> is a negative time range.</exception>
+        public static void LogDependency(
+            this ILogger logger,            
+            string dependencyType,
+            object dependencyData,
+            string targetName,
+            bool isSuccessful,
+            string dependencyName,
```
tomkerkhove, line 439: Ordering

src/Arcus.Observability.Telemetry.Serilog.Sinks.ApplicationInsights/Converters/OperationContextConverter.cs
```diff
@@ -31,5 +32,41 @@ public class OperationContextConverter
                 telemetryEntry.Context.Operation.ParentId = operationParentId;
             }
         }
+
+        /// <summary>
+        /// Enrich the given <paramref name="telemetryEntry"/> with the operation name.
+        /// </summary>
+        /// <param name="telemetryEntry">The telemetry instance to enrich.</param>
+        public void EnrichWithOperationName<TEntry>(TEntry telemetryEntry) where TEntry : ITelemetry, ISupportProperties
+        {
```
tomkerkhove, line 41: Just a nit on the naming, can you rename things such as `r` to `requestTelemetry` please?
We don't tend to use abbreviations

src/Arcus.Observability.Telemetry.Serilog.Sinks.ApplicationInsights/Converters/RequestTelemetryConverter.cs
```diff
@@ -42,9 +43,14 @@ protected override RequestTelemetry CreateTelemetryEntry(LogEvent logEvent, IFor
             var requestTelemetry = new RequestTelemetry(requestName, requestTime, requestDuration, responseStatusCode, isSuccessfulRequest)
             {
                 Id = operationId,
-                Url = url,
+                Url = url
             };
 
+            if (!String.IsNullOrEmpty(operationName))
```
tomkerkhove, line 49: Since we have the operation converter, do we still need this? Shouldn't we rely on the request telemetry object here?

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
pim-simons: I had it like that initially but then I run into:
```The call is ambiguous between the following methods or properties: 'ILoggerExtensions.LogDependency(ILogger, string, string, object, bool, DateTimeOffset, TimeSpan, Dictionary<string, object>)' and 'ILoggerExtensions.LogDependency(ILogger, string, object, string, bool, DateTimeOffset, TimeSpan, Dictionary<string, object>)'```

Specifically these methods are ambiguous:
```
public static void LogDependency(
	this ILogger logger,            
	string dependencyType,
	string dependencyName,
	object dependencyData,
	bool isSuccessful,
	DateTimeOffset startTime,
	TimeSpan duration,
	Dictionary<string, object> context = null)
```
and 
```
public static void LogDependency(
	this ILogger logger,
	string dependencyType,
	object dependencyData,
	string targetName,
	bool isSuccessful,
	DateTimeOffset startTime,
	TimeSpan duration,
	Dictionary<string, object> context = null)
```

I figured to put the dependencyName as the last value before the DateTimeOffset/DependencyMeasurement as this will make the methods unique. Any other options to consider here?

File: src/Arcus.Observability.Telemetry.Serilog.Sinks.ApplicationInsights/Converters/OperationContextConverter.cs (hunk 14 above)
Comment (pim-simons) [reply src/Arcus.Observability.Telemetry.Serilog.Sinks.ApplicationInsights/Converters/OperationContextConverter.cs:41]: Changed, see https://github.com/arcus-azure/arcus.observability/pull/229/commits/ece70efeabf862f09c7811c5ebae8ac38a275aa1

File: src/Arcus.Observability.Telemetry.Core/Logging/RequestLogEntry.cs (hunk 9 above)
Comment (pim-simons) [reply src/Arcus.Observability.Telemetry.Core/Logging/RequestLogEntry.cs:31]: Changed it so we call the other constructor and assing a default operation, see https://github.com/arcus-azure/arcus.observability/pull/229/commits/8804c4bea20c511e20149bd502a02fe8d8a65640

File: src/Arcus.Observability.Telemetry.Serilog.Sinks.ApplicationInsights/Converters/RequestTelemetryConverter.cs (hunk 15 above)
Comment (pim-simons) [reply src/Arcus.Observability.Telemetry.Serilog.Sinks.ApplicationInsights/Converters/RequestTelemetryConverter.cs]: I had to change the implementation here because we now set a default value in the RequestLogEntry constructor, see https://github.com/arcus-azure/arcus.observability/pull/229/commits/8804c4bea20c511e20149bd502a02fe8d8a65640
I now check if the request method is present in the operation name (in case a custom operation name is supplied), if not it is added.

I am setting the operation on the ```requestTelemetry.Context.Operation.Name```, but I am not sure if this solves your comment "Shouldn't we rely on the request telemetry object here?". I could use some pointers if certain functionality should be implemented differently

File: src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs (hunk 10 above)
Comment (tomkerkhove) [reply src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs:241]: Dammit! No problem!

File: src/Arcus.Observability.Telemetry.Core/Logging/RequestLogEntry.cs [hunk 16]
```diff
@@ -28,10 +28,44 @@ public class RequestLogEntry
...
+            string host,
+            string uri,
+            int statusCode,
+            TimeSpan duration,
+            IDictionary<string, object> context) : this (method, host, uri, $"{method} {uri}", statusCode, duration, context)
+        {
+            Guard.For<ArgumentException>(() => host?.Contains(" ") == true, "Requires a HTTP request host name without whitespace");
+            Guard.NotLessThan(statusCode, 100, nameof(statusCode), "Requires a HTTP response status code that's within the 100-599 range to track a HTTP request");
+            Guard.NotGreaterThan(statusCode, 599, nameof(statusCode), "Requires a HTTP response status code that's within the 100-599 range to track a HTTP request");
+            Guard.NotLessThan(duration, TimeSpan.Zero, nameof(duration), "Requires a positive time duration of the request operation");
```
tomkerkhove, line 42: Since we call the other constructor we can leave these out

src/Arcus.Observability.Telemetry.Core/Logging/RequestLogEntry.cs
```diff
@@ -46,6 +80,7 @@ public class RequestLogEntry
             RequestUri = uri;
             ResponseStatusCode = statusCode;
             RequestDuration = duration;
+            OperationName = operationName;
```
tomkerkhove, line 83: We should guard not null or whitespace here

tomkerkhove (approved):
LGTM, thank you! 🎉

src/Arcus.Observability.Telemetry.Core/Logging/RequestLogEntry.cs
pim-simons: Changed, see: https://github.com/arcus-azure/arcus.observability/pull/229/commits/7dbacede883ad62c749d89fcfc001c7d465891da

src/Arcus.Observability.Telemetry.Core/Logging/RequestLogEntry.cs
pim-simons: Added, see: https://github.com/arcus-azure/arcus.observability/pull/229/commits/4fb6cf03292fa42a0876bfc0d1ba5db4716c3554

tomkerkhove:
I think we are good to go, except for the integration tests - Are these on your radar?

pim-simons:
> 
> 
> I think we are good to go, except for the integration tests - Are these on your radar?

I updated the integration tests to check for the dependencyname and added some integration tests for the requestlogging to test specifying a custom operation name. 
Am I missing any?

tomkerkhove:
They seem to be failing, can you please have a look?

pim-simons:
> 
> 
> They seem to be failing, can you please have a look?

Worst argument ever, but works on my machine 😉 
image

I can't really find a reason why these specific tests would fail, is there any way to manually restart them?

pim-simons:
It's all looking green now 👍🏻 seems like the tests worked this time

tomkerkhove:
🙌 Thank you for the nice contribution!

Issue #231, opened 2021-09-07
Improve telemetry unit tests with strongly-typed written log message


stijnmoreels:
**Is your feature request related to a problem? Please describe.**
Currently, the only assertion we have in the telemetry logging tests, is to verify if the value exists somewhere in the log message. We don't assert on the format.

**Describe the solution you'd like**
A better way would be if we parse the written log message as a strongly-typed log entry so we get the initial type that was used to log the telemetry. This will make sure we are more sure of which values we assert on, and if these values are at the right place.
It will also make the unit tests more easy to understand.

**Additional context**
We could use regular expressions and capture groups to find more easily the necessary parts of the serialized log message.

stijnmoreels:
Done in combination with other work, no need to specify this explicitly as its spans milestones.

PR #232, opened 2021-09-09
test: strongly-typed metric assertion in unit tests

stijnmoreels:
Improvement on the current rather 'loose' assertion. This improvement will make testing easier to understand and more strict.

Relates to #231

### DIFF
--- a/src/Arcus.Observability.Tests.Unit/Extensions/TestLoggerExtensions.cs
+++ b/src/Arcus.Observability.Tests.Unit/Extensions/TestLoggerExtensions.cs
@@ -35,7 +35,7 @@ namespace Arcus.Observability.Tests.Unit
                     "Cannot parse the written message as a telemetry metric because no log message was written to this test logger");
             }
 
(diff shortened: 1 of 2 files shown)


src/Arcus.Observability.Tests.Unit/Extensions/TestLoggerExtensions.cs
```diff
@@ -0,0 +1,116 @@
...
+                           .ToDictionary(
+                               item => item[0].Trim(trailingChars),
+                               item => (object) item[1].Trim(trailingChars));
+
+            Assert.Single(context, item =>
+            {
+                return item.Key == ContextProperties.General.TelemetryType
+                       && item.Value.ToString() == TelemetryType.Metrics.ToString();
+            });
+            context.Remove(ContextProperties.General.TelemetryType);
```
stijnmoreels, line 111: The log entries already add this telemetry type by-default. If we don't remove it, it will be there twice. The assertion above this line will make sure that the telemetry type is there so we don't have to check this in the test.

tomkerkhove (approved):

PR #235, opened 2021-09-30
docs: We only use {major}.{minor} for doc versions

tomkerkhove:
Signed-off-by: Tom Kerkhove <kerkhove.tom@gmail.com>


stijnmoreels (approved):
Aha, that's true. Thx.

PR #250, opened 2021-10-26
feat: add az servicebus request tracking (w/o ai support)

### LINKED ISSUE #242: Provide support to track Azure Service Bus requests
Our `.LogRequest` functionality only takes into account HTTP requests and no other types. This is troublesome if we want to link Azure Service Bus resources to a service-to-service correlation.

We should provide support to track Azure Service Bus requests. This change will require some additions and changes:
* [x] Add logging extension `.LogAzureServiceBusRequest` specific for Azure Service Bus requests which should take in:
    * Azure Service Bus entity name
    * Azure Service Bus namespace
    * Operation name (default to `Process`)
* [x] Add `ServiceBus-Entity` in the telemetry context with the Azure Service Bus entity name
* [x] Add `ServiceBus-Endpoint` in the telemetry context with the following format:
        `type:Azure Service Bus | name:{entityName} | endpoint:sb://{namespaceEndpoint}.servicebus.windows.net/`
* [x] Use `Azure Service Bus` as `SourceSystem` of the request telemetry.
* [x] Use operation name as `SourceName` of the request telemetry.

> Note that HTTP-specific properties in the request telemetry should be filled-out, but not with anything important.

**Additional context**
[Azure Service Bus telemetry tracking in the official Application Insights SDK](https://github.com/microsoft/ApplicationInsights-dotnet/blob/3822ab1c591298b4c0c00eb6a853265a180e8d70/WEB/Src/DependencyCollector/DependencyCollector/Implementation/EventHandlers/ServiceBusDiagnosticsEventHandler.cs)
[Service-to-service poc](https://github.com/arcus-azure/arcus-service-to-service-correlation-poc)

stijnmoreels:
Adds request tracking for Azure Service Bus queue/topic/unknown without (yet) Application Insights support.

Closes #242


docs/preview/02-Features/writing-different-telemetry-types.md
```diff
@@ -429,6 +429,38 @@ logger.LogMetric("Invoice Received", 133.37, telemetryContext);
 
 ## Requests
 
+### Incoming Azure Service Bus requests
+Requests allow you to keep track of Azure Service Bus messages on a queue or topic and how the message was processed.
```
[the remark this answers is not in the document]
fgheysels:
```suggestion
Requests allow you to keep track of incoming Azure Service Bus messages on a queue or topic and how the message was processed.
```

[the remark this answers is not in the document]
fgheysels: Regarding the discussion on `topics` and `subscriptions`.   You define a topic in Azure Service Bus, and you define one (or more) subscriptions on that topic.   You receive the message through a subscription.

docs/preview/02-Features/writing-different-telemetry-types.md
```diff
@@ -429,6 +429,38 @@ logger.LogMetric("Invoice Received", 133.37, telemetryContext);
...
+
+Here is how you can report an Azure Service Bus queue request on a message that's being processed:
```
fgheysels, line 435: report or log ?

docs/preview/02-Features/writing-different-telemetry-types.md
```diff
@@ -429,6 +429,38 @@ logger.LogMetric("Invoice Received", 133.37, telemetryContext);
...
+    {
+        // Processing message.
+
+        // End processing.
+        
+        isSuccessful = true;
+    }
+    finally
+    {
+        logger.LogServiceBusQueueRequest("<my-queue-namespace>.servicebus.windows.net", "<my-queue-name>", "<operation-name>", isSuccessful, measurement);
```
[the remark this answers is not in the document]
fgheysels: I think @tomkerkhove  meant to pass in the namespace name of the ServiceBus, so without the `.servicebus.windows.net` part.

Actually, that would mean that the method signature would look like this:

```csharp
LogServiceBusQueueRequest( this ILogger logger,
    string namespacename,
    string queuename,
    string operationname,
    bool isSuccessful,
    DependencyMeasurement m,
    string namespaceSuffix = ".servicebus.windows.net
```
By doing that, you do not force developers to always add the '.servicebus.windows.net' part to their logs.  Moreover, I think they'll typically get the namespace-name from their Configuration (which probably doesn't contain that suffix).
If they must change the suffix, they can do it via the optional parameter.

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
```diff
@@ -189,6 +189,209 @@ public static class ILoggerExtensions
...
+            bool isSuccessful,
+            DependencyMeasurement measurement,
+            Dictionary<string, object> context = null)
+        {
+            Guard.NotNull(logger, nameof(logger), "Requires an logger instance to track telemetry");
+            Guard.NotNullOrWhitespace(serviceBusNamespace, nameof(serviceBusNamespace), "Requires an Azure Service Bus namespace to track the topic request");
+            Guard.NotNullOrWhitespace(topicName, nameof(topicName), "Requires an Azure Service Bus topic name to track the topic request");
+            Guard.NotNull(measurement, nameof(measurement), "Requires an instance to measure the Azure Service Bus topic request process latency duration");
+
+            // TODO: the dependency date from the measurement instance is not used when tracking the request.
```
fgheysels, line 218: Must something be done with this todo before completing the PR ?

src/Arcus.Observability.Telemetry.Core/ContextProperties.cs
```diff
@@ -64,6 +64,15 @@ public static class RequestTracking
...
             public const string RequestTime = "RequestTime";
+            
+            public static class ServiceBus
+            {
+                public const string Endpoint = "ServiceBus-Endpoint";
+                public const string EntityName = "ServiceBus-Entity";
+                public const string EntityType = "ServiceBus-EntityType";
+
+                public const string DefaultOperationName = "Process";
+            }
```
fgheysels, line 75: Nested public classes, that's a bit of a smell, don't you think ?
Do these classes need to be publicanyway ?  To me, it seems that they're only used internally ?  (Maybe I'm wrong)

docs/preview/02-Features/writing-different-telemetry-types.md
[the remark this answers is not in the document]
stijnmoreels: Best not to use optional parameters in public signatures of a library, as it's hard to maintain. We could create an overload for this, though.

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
stijnmoreels: Not sure, was something I came across. We don't use the data of the measurement, only the tracked time. It may not be an issue here, though.

src/Arcus.Observability.Telemetry.Core/ContextProperties.cs
stijnmoreels: The purpose of those nested classes is to have a kindoff namespace and get more organization on our constants. I don't consider it a smell in this case. 😄 There're used in our Application Insights package to get them back out, I guess. And also, to make it possible for consumers to write their own enrichers and have easy access to the telemetry keys used in the telemetry context.

docs/preview/02-Features/writing-different-telemetry-types.md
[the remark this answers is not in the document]
stijnmoreels: Yeah, ok, I got that, but not sure what has to change here 😅 .

src/Arcus.Observability.Telemetry.Core/ContextProperties.cs
fgheysels: Ok.

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
fgheysels: I guess we'll need to build a preview package , call this from a test project and see how it is displayed in Application Insights then ?

docs/preview/02-Features/writing-different-telemetry-types.md
```diff
@@ -429,6 +429,38 @@ logger.LogMetric("Invoice Received", 133.37, telemetryContext);
 
 ## Requests
 
+### Incoming Azure Service Bus requests
+Requests allow you to keep track of incoming Azure Service Bus messages on a queue or topic and how the message was processed.
```
fgheysels, line 433:
```suggestion
Requests allow you to keep track of incoming Azure Service Bus messages on a queue or topic.
```
I think we can remove the 'and how the message was processed' part, as that is no longer the responsability of the request that we're logging.

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
stijnmoreels: We can already do that in the integration tests, but this PR doesn't yet do the link with Application Insights (see description), so we can't test this yet.

stijnmoreels: I don't think this will have issues as the dependency data is used for dependency tracking, but not sure. Maybe we should include the data in the telemetry context.

docs/preview/02-Features/writing-different-telemetry-types.md
stijnmoreels: 😄 Not sure.

docs/preview/02-Features/writing-different-telemetry-types.md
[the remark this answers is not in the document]
stijnmoreels: Added methods with `...WithPrefix` to make this distinction more clear.  https://github.com/arcus-azure/arcus.observability/pull/250/commits/9bacc6bd8c21cfddaa2decde9a9360d61a156254

docs/preview/02-Features/writing-different-telemetry-types.md
fgheysels: I'd say log, but that's a minor thing :)

fgheysels:
```suggestion
Here is how you can log an Azure Service Bus queue request on a message that's being processed:
```

docs/preview/02-Features/writing-different-telemetry-types.md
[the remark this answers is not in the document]
fgheysels: It's a suffix, and not a prefix :)

fgheysels (approved):
I think this is a good start, and we can continue from here.
Just a few remarks though.

docs/preview/02-Features/writing-different-telemetry-types.md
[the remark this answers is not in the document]
stijnmoreels: Wel... you pass along the prefix, right, as in `{prefix}.servicebus.windows.net`. I remember a similar method overload in messaging where we do the same.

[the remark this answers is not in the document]
fgheysels: The 'prefix' is what is mandatory to pass; the suffix (.servicebus.windows.net) is optional, as that may vary between Azure Clouds.
For the global Azure Cloud, this is `.servicebus.windows.net`.    In the German Azure Cloud, this is `servicebus.cloudapi.de`, for the Azure cloud in China this is yet another address (`servicebus.chinacloudapi.cn`).
So , that suffix is the part that should be optional.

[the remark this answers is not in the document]
stijnmoreels: Yes, I got that, that's how we did it in messaging too. Hmm, I think we misunderstand each other. 😅 
Let's take a step back:

We have two type of methods, now. One that takes in the fully qualified name and one that only takes in the prefix.
These methods are called:
- `LogServiceBusRequest...()` 
- `LogServiceBusRequestWithPrefix...()`

This way we can take care of different clouds as well as giving consumers the possibility to manipulate the fully qualified name.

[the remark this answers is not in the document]
fgheysels: Now I see; I just think that having this method named `...WithPrefix` adds to the confusion.  I'd just create it as a regular overload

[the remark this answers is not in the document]
stijnmoreels: The problem is that we'll have signature conflicts, then, if we use the same method name. I used the same way of working as with Arcus Messaging, were we also include `...WithPrefix` in the method name when we want to provide a prefix instead of the full namespace. I don't see immediately a clearer solution besides renaming the method (like now). Moving the parameters around to get a different signature is probably not a very good developer experience and would maybe cause problems. With a different method name, we clearly state that we will use it differently.
Any suggestions on a different name, then? I wanted to be consistent but if it's causing confusion, then we could look for something else.

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
stijnmoreels: I'll add an issue for this and close this conv.

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
```diff
@@ -189,6 +189,402 @@ public static class ILoggerExtensions
...
+        /// <param name="logger">The logger instance to track the telemetry.</param>
+        /// <param name="serviceBusNamespace">The namespace where the Azure Service Bus topic is registered.</param>
+        /// <param name="topicName">The name of the Azure Service Bus topic.</param>
+        /// <param name="operationName">The name of the operation of the request.</param>
+        /// <param name="isSuccessful">The indication whether or not the Azure Service Bus topic request was successfully processed.</param>
+        /// <param name="measurement">The instance to measure the latency duration of the Azure Service Bus topic request.</param>
+        /// <param name="context">The telemetry context that provides more insights on the Azure Service Bus topic request.</param>
+        /// <exception cref="ArgumentNullException">Thrown when the <paramref name="logger"/> or the <paramref name="measurement"/> is <c>null</c>.</exception>
+        /// <exception cref="ArgumentException">Thrown when the <paramref name="serviceBusNamespace"/> or <paramref name="topicName"/> is blank.</exception>
+        public static void LogServiceBusTopicRequest(
```
fgheysels, line 235: So, related to my previous comment, this method would be named `LogServiceBusTopicRequestWithSuffix` and this method introduces an extra parameter where you can specify the 'cloud' that is used.

It would then be called like this:

```
logger.LogServiceBusTopicRequestWithSuffix ("mynamespacename", suffix: ".servicebus.cloudapi.de", topicName: "mytopic", .... );
```

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
```diff
@@ -189,6 +189,402 @@ public static class ILoggerExtensions
...
+        /// <exception cref="ArgumentOutOfRangeException">Thrown when the <paramref name="duration"/> is a negative time range.</exception>
+        public static void LogServiceBusQueueRequest(
+            this ILogger logger,
+            string serviceBusNamespace,
+            string queueName,
+            string operationName,
+            bool isSuccessful,
+            TimeSpan duration,
+            DateTimeOffset startTime,
+            Dictionary<string, object> context = null)
```
fgheysels, line 434: Just asking: do we need to make a distinction between topics and queues here ?  

If so, then I'd suggest that we extend the 'TopicRequest' method signatures with a `subscriptionName` parameter, as you receive messages that are posted on a topic through a subscription.

src/Arcus.Observability.Telemetry.Core/Logging/RequestSourceSystem.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace Arcus.Observability.Telemetry.Core.Logging
+{
+    /// <summary>
+    /// Represents the system from where the request came from.
+    /// </summary>
+    public enum RequestSourceSystem
+    {
+        /// <summary>
+        /// Sets the request source system as an Azure Service Bus queue or topic.
```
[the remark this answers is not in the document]
stijnmoreels: Yes, indeed. Thx!

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
```diff
@@ -189,6 +189,402 @@ public static class ILoggerExtensions
...
+        /// <param name="serviceBusNamespacePrefix">The namespace prefix where the Azure Service Bus topic is registered (use Azure public cloud).</param>
+        /// <param name="topicName">The name of the Azure Service Bus topic.</param>
+        /// <param name="operationName">The name of the operation of the request.</param>
+        /// <param name="isSuccessful">The indication whether or not the Azure Service Bus topic request was successfully processed.</param>
+        /// <param name="measurement">The instance to measure the latency duration of the Azure Service Bus topic request.</param>
+        /// <param name="context">The telemetry context that provides more insights on the Azure Service Bus topic request.</param>
+        /// <exception cref="ArgumentNullException">Thrown when the <paramref name="logger"/> or the <paramref name="measurement"/> is <c>null</c>.</exception>
+        /// <exception cref="ArgumentException">Thrown when the <paramref name="serviceBusNamespacePrefix"/> or <paramref name="topicName"/> is blank.</exception>
+        public static void LogServiceBusTopicRequestWithPrefix(
```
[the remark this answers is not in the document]
stijnmoreels: Ok, that would make sense. Let's do that!

src/Arcus.Observability.Telemetry.Core/Logging/RequestLogEntry.cs
```diff
@@ -64,24 +70,74 @@ public class RequestLogEntry
...
+
+        private RequestLogEntry(
+            string method, 
+            string host, 
+            string uri, 
+            string operationName,
+            int statusCode,
+            RequestSourceSystem sourceSystem,
+            TimeSpan duration,
+            string requestTime,
```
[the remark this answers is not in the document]
stijnmoreels: Hihi, we already do that. You see, this ctor is `private` and is called from `static` methods where we do the conversion from `DateTimeOffset` to `string` so we get a consistent `string` representation of this. We do need it as a `string` bc these types will be used as serialization, and with this, we can control how the `DateTimeOffset` is serialized (see `FormatSpecifiers.InvariantTimestampFormat`).

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
stijnmoreels: Normally, this is already available in the Azure SDK, but it seems unnecessary to reference this. So yeah, I think the best way in between is to use the `string` representation here. People can still reference the Azure SDK themselves, then.

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
stijnmoreels: We do a distinction on the `ServiceBusEntityType`, yeah. Ah, indeed. An extra parameter for this would be nice. Thx!

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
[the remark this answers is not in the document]
stijnmoreels: I think Tom wanted the other thing, where you could only specify the first part (prefix) of the namespace, without worrying about the remaining Azure Cloud.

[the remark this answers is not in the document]
stijnmoreels: Changed in https://github.com/arcus-azure/arcus.observability/pull/250/commits/8d1822fa0d64c719ae50d97beb6de3ca32fab666

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
stijnmoreels: Added topic subscription in https://github.com/arcus-azure/arcus.observability/pull/250/commits/361b9d0c65a3c438165e78a5030282fa757dea40

src/Arcus.Observability.Telemetry.Core/Logging/RequestLogEntry.cs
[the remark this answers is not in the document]
fgheysels: I just thought it would be best to do that as late as possible, and thus, pass it as a datetimeoffset to the private ctor, and do the 'ToString' stuff inside the ctor.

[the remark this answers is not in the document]
stijnmoreels: Hihi, yeah, I think this is this kind of style where you put no functionality in ctors and only assignments. I think we did that around Arcus that way, but we can check.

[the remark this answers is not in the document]
fgheysels: I'm OK with it like it is for now.   It's not that it will be a breaking change should we ever refactor this.
IMHO, it's just best do to this as late as possible.

stijnmoreels:
> MIght be good if we just add an example of what is meant with that suffix ?

Yes, that would be good. I add it to the XML docs, and the feature docs.

docs/preview/02-Features/writing-different-telemetry-types.md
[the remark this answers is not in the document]
fgheysels: It seems like there was a difference in interpretation; we took this offline.

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
```diff
@@ -189,6 +189,444 @@ public static class ILoggerExtensions
             logger.LogWarning(RequestFormat, new RequestLogEntry(request.Method.ToString(), host, resourcePath, operationName, statusCode, duration, context));
         }
 
+        /// <summary>
+        ///     Logs an Azure Service Bus topic request.
+        /// </summary>
+        /// <param name="logger">The logger instance to track the telemetry.</param>
+        /// <param name="serviceBusNamespace">The namespace where the Azure Service Bus topic is registered. (Will be combined with <paramref name="serviceBusNamespaceSuffix"/>).</param>
+        /// <param name="serviceBusNamespaceSuffix">The namespace suffix (i.e. '.servicebus.windows.net') where the Azure Service Bus topic is registered (Will be combined with <paramref name="serviceBusNamespace"/>).</param>
```
fgheysels, line 197:
```suggestion
        /// <param name="serviceBusNamespaceSuffix">The namespace suffix (i.e. '.servicebus.windows.net' or '*.servicebus.cloudapi.de') where the Azure Service Bus topic is registered (Will be combined with <paramref name="serviceBusNamespace"/>).</param>
```

fgheysels: I'd add an extra example

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
fgheysels: I agree.  It would be nice if we could just mention this in the documentation.

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
```diff
@@ -189,6 +189,444 @@ public static class ILoggerExtensions
...
+        }
+
+        /// <summary>
+        ///     Logs an Azure Service Bus queue request.
+        /// </summary>
+        /// <param name="logger">The logger instance to track the telemetry.</param>
+        /// <param name="serviceBusNamespace">The namespace where the Azure Service Bus queue is registered. (Will be combined with <paramref name="serviceBusNamespaceSuffix"/>).</param>
+        /// <param name="serviceBusNamespaceSuffix">The namespace suffix (i.e. '.servicebus.windows.net') where the Azure Service Bus queue is registered (Will be combined with <paramref name="serviceBusNamespace"/>).</param>
+        /// <param name="queueName">The name of the Azure Service Bus queue.</param>
+        /// <param name="operationName">The name of the operation of the request.</param>
```
fgheysels, line 355:
```suggestion
        /// <param name="operationName">An optional logical name that can be used to identify the operation that consumes the message.</param>
```

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
```diff
@@ -189,6 +189,444 @@ public static class ILoggerExtensions
...
+        /// <exception cref="ArgumentNullException">Thrown when the <paramref name="logger"/> or the <paramref name="measurement"/> is <c>null</c>.</exception>
+        /// <exception cref="ArgumentException">Thrown when the <paramref name="serviceBusNamespace"/> or <paramref name="entityName"/> is blank.</exception>
+        public static void LogServiceBusRequestWithSuffix(
+            this ILogger logger,
+            string serviceBusNamespace,
+            string serviceBusNamespaceSuffix,
+            string entityName,
+            string operationName,
+            bool isSuccessful,
+            DependencyMeasurement measurement,
```
fgheysels, line 501: Should we create a `RequestMeasurement` instead ?  Or can we make the `DependencyMeasurement` type more generic so that it can be renamed to `Measurement` ?

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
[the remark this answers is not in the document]
fgheysels: Great, I think this is clear now!

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
fgheysels: Created an issue for this. #267

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
fgheysels: Created an issue for this #268

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
stijnmoreels: Done in https://github.com/arcus-azure/arcus.observability/pull/250/commits/b1f36d60ca96e8b59a2bb79bc1116c270a8380c1

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
stijnmoreels: Done in https://github.com/arcus-azure/arcus.observability/pull/250/commits/3b9209236262c1cc2fee6f4f4c99efc8c4548795

PR #257, opened 2021-12-10
chore: upgrade to net6.0

stijnmoreels:
Upgrades towards .NET 6 while keeping .NET Core backwards compability.

Relates to https://github.com/arcus-azure/arcus/issues/106


stijnmoreels:
One can argue that we should run the tests twice (for both frameworks)...

fgheysels (approved):

PR #262, opened 2021-12-31
fix: use serilog scalar value when serializing string values

### LINKED ISSUE #256: Correctly support scalar and other Serilog log property values when parsing to `string`
**Describe the bug**
When converting the Serilog values to Azure Application Insights models, we have a `.ToDecentString` method that takes in the Serilog log property value and returns the simple string representation of that value.
This works for simple values, but more complex values (like JSON values), this will result in escaped characters.

**To Reproduce**
Add a telemetry context to any logging extension method with a JSON value.

**Expected behavior**
The end-result Azure Application Insights telemetry value should contain the original JSON value (unescaped).

**Additional context**
 - Current version (v2.3.0)
 - Relates to https://github.com/arcus-azure/arcus.webapi/issues/270

stijnmoreels:
Use the Serilog `ScalarValue` (represents simple value) when serializing the Serilog log property value to a `string` representation. This fix will make sure that we retrieve the original value (unescaped). This will make JSON values represents as their original values instead of escaped quotes (`\"`).

Closes #256

### DIFF
--- a/src/Arcus.Observability.Telemetry.Serilog.Enrichers/Extensions/LogEventPropertyValueExtensions.cs
+++ b/src/Arcus.Observability.Telemetry.Serilog.Enrichers/Extensions/LogEventPropertyValueExtensions.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Globalization;
+using System.IO;
 using System.Linq;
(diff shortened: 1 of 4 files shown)


fgheysels (approved):
Great, thx for fixing this \o/

PR #270, opened 2022-01-21
feat: add request measurement for requests

### LINKED ISSUE #268: Introduce a `RequestMeasurement` which is used to log the duration of a request
When logging a ServiceBus or HTTP request, we now pass in a `DependencyMeasurement` type as a parameter.

We should introduce a `RequestMeasurement` for these requests instead.

Once this type is available, we must use this in the  LogServiceBus and LogHttp request methods.

_Originally posted by @fgheysels in https://github.com/arcus-azure/arcus.observability/pull/250#discussion_r788505595_

stijnmoreels:
Use a new `RequestMeasurement` object to measure requests.

Closes #268

### DIFF
--- a/docs/preview/02-Features/writing-different-telemetry-types.md
+++ b/docs/preview/02-Features/writing-different-telemetry-types.md
@@ -330,7 +330,7 @@ logger.LogDependency("SendGrid", dependencyData, isSuccessful: true, startTime: 
 // Output: {"DependencyType": "SendGrid", "DependencyData": "http://my.sendgrid.uri/", "Duration": "00:00:01.2396312", "StartTime": "03/23/2020 09:32:02 +00:00", "IsSuccessful": true, "Context": {}}
 ```
 
-### Making it easiser to measure telemetry
+### Making it easier to measure telemetry
 
 Measuring dependencies or requests means you need to keep track of how long the action took and when it started.
 The `Arcus.Observability.Telemetry.Core` library provides an easy way to accomplish this.
--- a/docs/preview/02-Features/writing-different-telemetry-types.md
+++ b/docs/preview/02-Features/writing-different-telemetry-types.md
@@ -330,9 +330,10 @@ logger.LogDependency("SendGrid", dependencyData, isSuccessful: true, startTime: 
 // Output: {"DependencyType": "SendGrid", "DependencyData": "http://my.sendgrid.uri/", "Duration": "00:00:01.2396312", "StartTime": "03/23/2020 09:32:02 +00:00", "IsSuccessful": true, "Context": {}}
 ```
 
-### Making it easier to measure dependencies
+### Making it easiser to measure telemetry
 
-Measuring dependencies means you need to keep track of how long the action took and when it started.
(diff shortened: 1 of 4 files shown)


stijnmoreels:
Point of notice: the current `.LogRequest` and its overloads don't use the `DepenencyMeasurement`, so no obsolete-making was necessary. They use a single `TimeSpan` for duration and are using the `DateTimeOffset.UtcNow` internally as start time.

docs/preview/02-Features/writing-different-telemetry-types.md
```diff
@@ -330,9 +330,10 @@ logger.LogDependency("SendGrid", dependencyData, isSuccessful: true, startTime:
 // Output: {"DependencyType": "SendGrid", "DependencyData": "http://my.sendgrid.uri/", "Duration": "00:00:01.2396312", "StartTime": "03/23/2020 09:32:02 +00:00", "IsSuccessful": true, "Context": {}}
 ```
 
-### Making it easier to measure dependencies
+### Making it easiser to measure telemetry
```
fgheysels, line 333:
```suggestion
### Making it easier to measure telemetry
```
small typo :)

fgheysels (commented):
Great, nice implementation & docs!

I just wonder why the `LogHttpRequest` methods do take a `Timespan` as argument instead of a `RequestMeasurement` ?

stijnmoreels:
> Great, nice implementation & docs!
> 
> I just wonder why the `LogHttpRequest` methods do take a `Timespan` as argument instead of a `RequestMeasurement` ?

They use `DateTimeOffset.UtcNow` internally as `startTime`. This is more a convenience here, bc it allows users not to worry about this. These extensions are for example used in the request middleware of our Web API where can directly call this, and it uses the current time as start time.
Other than Azure Service Bus (where you have to peek/receive the message yourself), this start time can be a little different, so we allow consumers to determine themselves when it should be tracked.

But you're actually right. We could have additional extension methods that takes in this measurement. But since there's nobody yet asking about it, I'm not sure if we should. On the other hand, one could argue if we should then include it here 😉 .

fgheysels:
> > Great, nice implementation & docs!
> > I just wonder why the `LogHttpRequest` methods do take a `Timespan` as argument instead of a `RequestMeasurement` ?
> 
> They use `DateTimeOffset.UtcNow` internally as `startTime`. This is more a convenience here, bc it allows users not to worry about this. These extensions are for example used in the request middleware of our Web API where can directly call this, and it uses the current time as start time. Other than Azure Service Bus (where you have to peek/receive the message yourself), this start time can be a little different, so we allow consumers to determine themselves when it should be tracked.
> 
> But you're actually right. We could have additional extension methods that takes in this measurement. But since there's nobody yet asking about it, I'm not sure if we should. On the other hand, one could argue if we should then include it here 😉 .

I was just thinking: why not being consistent and use the measurements types there as well.   But if you have a good reason we can leave it as is.   If we have a request for this, we can add it indeed.

fgheysels (approved):

PR #285, opened 2022-02-02
feat: add duration measurement io dependency measurement

### LINKED ISSUE #283: Create `DurationMeasurement` as alternative for the `DependencyMeasurement`
Currently, we use a separate `DependencyMeasurement` type to track the duration of dependency telemetry. This measurement type contains dependency data that isn't used while tracking the dependency telemetry.

Rename the `RequestMeasurement` to `DurationMeasurement` to be as alternative for the `DependencyMeasurement` which doesn't contain any dependency data.

**Additional context**
Discussed in #279

stijnmoreels:
Rename `RequestMeasurement` to `DurationMeasurement` and use that instead of `DependencyMeasurement`.

Closes #283

### DIFF
--- a/src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
+++ b/src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
@@ -1386,6 +1386,7 @@ namespace Microsoft.Extensions.Logging
         /// <param name="measurement">The measuring the latency to call the Service Bus dependency.</param>
         /// <param name="context">The context that provides more insights on the dependency that was measured.</param>
         /// <exception cref="ArgumentNullException">Thrown when the <paramref name="logger"/> or <paramref name="measurement"/> is <c>null</c>.</exception>
+        /// <exception cref="ArgumentException">Thrown when the <paramref name="queueName"/> is blank.</exception>
         public static void LogServiceBusQueueDependency(
             this ILogger logger,
             string queueName,
--- a/src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
+++ b/src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
@@ -24,7 +24,7 @@ namespace Microsoft.Extensions.Logging
                                       SecretNameRegex = new Regex(SecretNamePattern, RegexOptions.Compiled);
 
         /// <summary>
-        ///     Logs an HTTP request
+        /// Logs an HTTP request
         /// </summary>
         /// <param name="logger">The logger to track the telemetry.</param>
         /// <param name="request">Request that was done</param>
@@ -62,7 +62,7 @@ namespace Microsoft.Extensions.Logging
         }
 
         /// <summary>
-        ///     Logs an HTTP request
+        /// Logs an HTTP request
         /// </summary>
         /// <param name="logger">The logger to track the telemetry.</param>
         /// <param name="request">Request that was done</param>
@@ -102,7 +102,7 @@ namespace Microsoft.Extensions.Logging
         }
 
         /// <summary>
-        ///     Logs an HTTP request
+        /// Logs an HTTP request
         /// </summary>
         /// <param name="logger">The logger to track the telemetry.</param>
         /// <param name="request">Request that was done</param>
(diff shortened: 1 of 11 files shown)


src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
```diff
@@ -925,9 +1050,45 @@ public static class ILoggerExtensions
...
+            Guard.NotNull(logger, nameof(logger), "Requires an logger instance to write the Azure Key Vault dependency");
+            Guard.NotNullOrWhitespace(vaultUri, nameof(vaultUri), "Requires a non-blank URI for the Azure Key Vault");
+            Guard.NotNullOrWhitespace(secretName, nameof(secretName), "Requires a non-blank secret name for the Azure Key Vault");
+            Guard.NotNull(measurement, nameof(measurement), "Requires a dependency measurement instance to track the latency of the Azure Key Vault when tracking an Azure Key Vault dependency");
+            Guard.For<FormatException>(
+                () => !SecretNameRegex.IsMatch(secretName),
+                "Requires a Azure Key Vault secret name in the correct format, see https://docs.microsoft.com/en-us/azure/key-vault/general/about-keys-secrets-certificates#objects-identifiers-and-versioning");
+            Guard.For<UriFormatException>(
+                () => !KeyVaultUriRegex.IsMatch(vaultUri),
+                "Requires the Azure Key Vault host to be in the right format, see https://docs.microsoft.com/en-us/azure/key-vault/general/about-keys-secrets-certificates#objects-identifiers-and-versioning");
```
stijnmoreels, line 1084: If we already make a decision on whether or not to remove this, we can also remove it here. If we leave that for another version, we should be consistent and add them here (like it is now).

src/Arcus.Observability.Telemetry.Sql/Extensions/ILoggerExtensions.cs
```diff
@@ -32,21 +40,62 @@ public static void LogSqlDependency(this ILogger logger, string connectionString
...
+            string operationName,
+            DateTimeOffset startTime,
+            TimeSpan duration,
+            bool isSuccessful,
+            Dictionary<string, object> context = null)
         {
+            Guard.NotNull(logger, nameof(logger), "Requires a logger instance to track telemetry");
             Guard.NotNullOrEmpty(connectionString, nameof(connectionString));
-            var connection = new SqlConnectionStringBuilder(connectionString);
+            Guard.NotNullOrWhitespace(operationName, nameof(operationName), "Requires a non-blank name of the SQL operation to track a SQL dependency");
```
fgheysels, line 96: Am I mistaken or did we decide a while ago that we wouldn't guard on the operation-name ?  In other words, the `operationName` is not mandatory ?

fgheysels: See this discussion: https://github.com/arcus-azure/arcus.observability/discussions/265

It is about the `tableName` parameter, but I'll create another issue for this.

fgheysels (approved):
Looks good to me! thx.

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
fgheysels: I've created an issue for this: https://github.com/arcus-azure/arcus.observability/issues/287
We can remove them in another PR and include that change in v2.4.0

src/Arcus.Observability.Telemetry.Sql/Extensions/ILoggerExtensions.cs
stijnmoreels: > Am I mistaken or did we decide a while ago that we wouldn't guard on the operation-name ? In other words, the operationName is not mandatory ?

Yeah, but we still guard it now, so I was trying to be consistent here. (I added this guard, bc it was already guarding on the delegated `.LogSqlDependency` call, see: https://github.com/arcus-azure/arcus.observability/blob/612335d8a76b0455ae97380cc68d8f0b83166c67/src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs#L1780)

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
stijnmoreels: 👍 Ok, let's do it!

src/Arcus.Observability.Telemetry.Sql/Extensions/ILoggerExtensions.cs
fgheysels: I've created a separate issue, so this can then be removed in the PR that tackles that issue.

PR #318, opened 2022-02-07
fix: discard `/docs` folder

stijnmoreels:
Seems that the wildcard doesn't work, let's try this one.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -16,7 +16,7 @@
     "enabled": true
   },
    "ignorePaths": [
-    "/docs/**"
+    "**/docs/**"
   ],
   "vulnerabilityAlerts": {
     "enabled": true


.github/renovate.json
```diff
@@ -16,7 +16,7 @@
     "enabled": true
   },
    "ignorePaths": [
-    "/docs/**"
+    "**/docs/**"
```
fgheysels, line 19: Weird, as docs is in the root folder, is n't it

fgheysels (approved):

.github/renovate.json
stijnmoreels: Yes indeed. 😅 Was also thinking about the heading `/` maybe that's the problem. Not sure.

Issue #336, opened 2022-02-15
Traces, logs, correlation are lost when using Arcus.Observability in Azure Functions


fgheysels:
**Describe the bug**

I have a Timer Triggered Azure Function, in where I use Arcus.Observability (and thus also Serilog) to log to Azure Application Insights.

Although there are traces being logged to Application Insights, they're not correlated.   With that , I mean that I cannot get an overview of traces (dependencies, etc... ) that were logged during one Function Invocation.

When I do not use Arcus.Observability and Serilog, everything is nicely linked AND there are things logged that are not logged out-of-the box when using Arcus.

**To Reproduce**

This is how I setup logging using Arcus:

```csharp
internal class Startup : FunctionsStartup
{
        public override void Configure(IFunctionsHostBuilder builder)
        {
            var configuration = builder.GetContext().Configuration;

            builder.ConfigureSecretStore(secretStoreBuilder => AddSecretProviders(secretStoreBuilder, configuration));

            builder.Services.AddLogging(loggingBuilder => ConfigureLogging(loggingBuilder, configuration));
        }

        private static void ConfigureLogging(ILoggingBuilder builder, IConfiguration configuration)
        {
            var functionDependencyContext = DependencyContext.Load(typeof(Startup).Assembly);
            
            var logConfiguration = new LoggerConfiguration()
                                        .MinimumLevel.Is(LogEventLevel.Information)
                                        .Enrich.FromLogContext()                                           
                                        .Enrich.WithComponentName("curated-dsbuilder-plc")
                                        .WriteTo.Console();

            var telemetryKey = configuration.GetValue<string>("APPINSIGHTS_INSTRUMENTATIONKEY");

            if (!String.IsNullOrWhiteSpace(telemetryKey))
            {                
                logConfiguration.WriteTo.AzureApplicationInsights(telemetryKey);
            }

            builder.ClearProvidersExceptFunctionProviders();
            builder.AddSerilog(logConfiguration.CreateLogger());
        }
}
```
When the Function runs, I can find all sort of things in App Insights, but correlation info is missing.   The traces do not contain any `Operation Id` , `Parent Id` header
![image](https://user-images.githubusercontent.com/3605786/154086797-82badef5-0641-444d-acfa-11e714e2cb9d.png)

If I change the function, so that Arcus.Observability and the AppInsights extension is not used for logging, traces are correlated AND I see things being logged that are not logged when using Arcus.Observability:

[...]

PR #340, opened 2022-02-24
chore: Mark ClearProvidersExceptFunctionProviders as obsolete

fgheysels:
This method is not correctly implemented and calling this method has the side-effect that logging to Application Insights is not working correctly.
It is advised to not longer use this method.

See also #336

### DIFF
--- a/src/Arcus.Observability.Telemetry.AzureFunctions/Extensions/ILoggerBuilderExtensions.cs
+++ b/src/Arcus.Observability.Telemetry.AzureFunctions/Extensions/ILoggerBuilderExtensions.cs
@@ -1,4 +1,5 @@
-﻿using GuardNet;
+﻿using System;
+using GuardNet;
 using Microsoft.Extensions.DependencyInjection;
 
 // ReSharper disable once CheckNamespace
@@ -15,10 +16,11 @@ namespace Microsoft.Extensions.Logging
         /// except the specific Azure Functions registrations.
         /// </summary>
         /// <param name="loggingBuilder">The builder containing the <see cref="ILoggerProvider"/> registrations.</param>
+        [Obsolete("Calling this method causes issues with correctly writing log-information to Application Insights")]
         public static ILoggingBuilder ClearProvidersExceptFunctionProviders(this ILoggingBuilder loggingBuilder)
         {
             Guard.NotNull(loggingBuilder, nameof(loggingBuilder));
-
+            
             // Kudos to katrash: https://stackoverflow.com/questions/45986517/remove-console-and-debug-loggers-in-asp-net-core-2-0-when-in-production-mode
             foreach (ServiceDescriptor serviceDescriptor in loggingBuilder.Services)
             {
--- a/src/Arcus.Observability.Telemetry.AzureFunctions/Extensions/ILoggerBuilderExtensions.cs
+++ b/src/Arcus.Observability.Telemetry.AzureFunctions/Extensions/ILoggerBuilderExtensions.cs
@@ -16,7 +16,7 @@ namespace Microsoft.Extensions.Logging
         /// except the specific Azure Functions registrations.
         /// </summary>
         /// <param name="loggingBuilder">The builder containing the <see cref="ILoggerProvider"/> registrations.</param>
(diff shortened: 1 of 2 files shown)


src/Arcus.Observability.Telemetry.AzureFunctions/Extensions/ILoggerBuilderExtensions.cs
```diff
@@ -15,6 +16,7 @@ public static class ILoggingBuilderExtensions
         /// except the specific Azure Functions registrations.
         /// </summary>
         /// <param name="loggingBuilder">The builder containing the <see cref="ILoggerProvider"/> registrations.</param>
+        [Obsolete("Calling this method causes issues with correctly writing log-information to Application Insights")]
```
stijnmoreels, line 19: Maybe something about not using it? 'Call this method on your own risk...' or something.

stijnmoreels (approved):
Yes, thx. 👍

src/Arcus.Observability.Telemetry.AzureFunctions/Extensions/ILoggerBuilderExtensions.cs
fgheysels: Yes, that would be good.   I added it.

src/Arcus.Observability.Tests.Unit/Correlation/DefaultCorrelationInfoAccessorTests.cs
```diff
@@ -19,15 +19,15 @@ public async Task SetCorrelationInfo_Twice_UsesMostRecentValue()
...
+            SetCorrelationInfo(secondOperationId, transactionId);
 
             // Assert
             CorrelationInfo correlationInfo = DefaultCorrelationInfoAccessor.Instance.GetCorrelationInfo();
             Assert.Equal(secondOperationId, correlationInfo.OperationId);
             Assert.Equal(transactionId, correlationInfo.TransactionId);
         }
 
-        private async Task SetCorrelationInfo(string operationId, string transactionId)
+        private void SetCorrelationInfo(string operationId, string transactionId)
```
stijnmoreels, line 30: Ah, no we can't do that. We should make this `async`  bc that was the cause why a previous bug was reported and this test checks that.

stijnmoreels (changes requested):
We should revert that last change of async method, bc that was a test for a previously reported bug.

src/Arcus.Observability.Tests.Unit/Correlation/DefaultCorrelationInfoAccessorTests.cs
fgheysels: I don't understand; what exactly are we testing ?  This method is marked as async, but doesn't do anything async ?

stijnmoreels: In a previous version, the `DefaultCorrelationAccessor` was not fully implemented correctly. It was not using the most recent set correlation information. This was caused due to asynchronous calls towards the accessor. The setup of the test was to verify that this is now fixed, and that asynchronous calls towards this accessor (and setting the correlation information) still results in the most recent set correlation information. 

If you want, you can add a task delay to make it more async, but we still need to make sure this is async, so it has to be reverted.

fgheysels: I will revert it, but to be honest, I don't think it will matter much. 
Since there are no awaitable calls, the method will just run synchronously.

fgheysels: Also, `DefaultCorrelationInfoAccessor.Instance` is deprecated ?

stijnmoreels: Yeah, I think it's mostly about context switching, not about asynchronous call explicitly.
Thx!
Yeah, indeed, so we probably will be able to remove the test once it's removed.

stijnmoreels (approved):

Issue #243, opened 2021-10-12
Provide the dependency ID when tracking dependencies


stijnmoreels:
**Is your feature request related to a problem? Please describe.**
When tracking dependencies, we should include an dependency ID so the later request tracking can use it for its parent operation ID in a service-to-service correlation scenario.

**Describe the solution you'd like**
We should include an `dependencyId` to all our available dependencies. This will usually result in method extension overloads.

**Additional context**
https://github.com/arcus-azure/arcus-service-to-service-correlation-poc

The dependency ID should be added to the following dependency types as extensions overloads:
- [x] Azure Key Vault (@stijnmoreels, #251 )
- [x] Azure search (@stijnmoreels, #345 )
- [x] Azure Service Bus Queue (@stijnmoreels #348 )
- [x] Azure Service Bus Topic (@stijnmoreels #360)
- [x] Azure Service Bus (general) (@stijnmoreels #348 )
- [x] Azure Blob storage (@stijnmoreels #370)
- [x] Azure Table storage (@stijnmoreels #371)
- [x] Azure Event Hubs (@stijnmoreels #378)
- [x] IoT Hub (@stijnmoreels #379 )
- [x] Cosmos SQL (@stijnmoreels #380)
- [x] HTTP dependency (@stijnmoreels #373)
- [x] SQL dependency (@stijnmoreels #376 )
- [x] Custom dependency (@stijnmoreels #362)

stijnmoreels:
Let's keep it as it is, for now. The signature is already complex enough, no need to further complicate it.

stijnmoreels:
The dependency ID should be added to the following dependency types as extensions overloads:
- [x] Azure Key Vault (@stijnmoreels, #251)
- [ ] Azure search
- [ ] Azure Service Bus Queue
- [ ] Azure Service Bus Topic
- [ ] Azure Service Bus (general)
- [ ] Azure Blob storage
- [ ] Azure Table storage
- [ ] Azure Event Hubs
- [ ] IoT Hub
- [ ] Cosmos SQL
- [ ] HTTP dependency
- [ ] SQL dependency
- [ ] Custom dependency

fgheysels:
I'm not sure if I'm following the use case here.   Maybe we can have a short call on this @stijnmoreels  ?

fgheysels:
@stijnmoreels  do we continue with this ?
In my opinion, HttpDependency should have priority here.

stijnmoreels:
[...]

PR #345, opened 2022-03-01
feat: add dependency id to az search

stijnmoreels:
Adds the capability to add a dependency ID when tracking Azure Search dependencies.

Related to #243

### DIFF
--- a/src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
+++ b/src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
@@ -1276,6 +1276,37 @@ namespace Microsoft.Extensions.Logging
         /// Logs an Azure Search Dependency.
         /// </summary>
         /// <param name="logger">The logger to track the telemetry.</param>
+        /// <param name="searchServiceName">The name of the Azure Search service.</param>
+        /// <param name="operationName">The name of the operation to execute on the Azure Search service.</param>
+        /// <param name="isSuccessful">The indication whether or not the operation was successful.</param>
+        /// <param name="measurement">The measuring the latency to call the dependency.</param>
+        /// <param name="dependencyId">The ID of the dependency to link as parent ID.</param>
+        /// <param name="context">The context that provides more insights on the dependency that was measured.</param>
+        /// <exception cref="ArgumentNullException">Thrown when the <paramref name="logger"/> or <paramref name="measurement"/> is <c>null</c>.</exception>
+        /// <exception cref="ArgumentException">Thrown when the <paramref name="searchServiceName"/> or <paramref name="operationName"/> is blank.</exception>
+        public static void LogAzureSearchDependency(
+            this ILogger logger,
+            string searchServiceName,
+            string operationName,
+            bool isSuccessful,
+            DurationMeasurement measurement,
+            string dependencyId,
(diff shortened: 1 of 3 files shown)


src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerExtensions.cs
```diff
@@ -1272,6 +1272,37 @@ public static class ILoggerExtensions
...
+        /// <param name="context">The context that provides more insights on the dependency that was measured.</param>
+        /// <exception cref="ArgumentNullException">Thrown when the <paramref name="logger"/> or <paramref name="measurement"/> is <c>null</c>.</exception>
+        /// <exception cref="ArgumentException">Thrown when the <paramref name="searchServiceName"/> or <paramref name="operationName"/> is blank.</exception>
+        public static void LogAzureSearchDependency(
+            this ILogger logger,
+            string searchServiceName,
+            string operationName,
+            bool isSuccessful,
+            DurationMeasurement measurement,
+            string dependencyId,
```
fgheysels, line 1293: So `dependencyId` is used to link this dependency to a certain parent ?
I assume that the name of this parameter is consistent with the name of related parameters in other methods, but I wonder if we can improve the name of the parameter to make this more clear ?
If this parameter is used to pass in the Id of the parent, so that we can link it to the parent, shouldn't we call it `parentId` ?

stijnmoreels: It's actually the way around, the ID is to link a request to the dependency. So, we have 'outgoing dependencies' and 'incoming requests'. This ID will be the parent ID of the request tracking. So, when tracking requests, it's set as the parent ID. In here, with dependencies, it's only to identify the dependency.

fgheysels: I don't understand.  Maybe we should discuss this offline.

stijnmoreels: Yeah, sure. It's something that we already discussed twice in separate meetings, 😅 . I hope when we get through with the service-to-service correlation, we can create some kind of diagram to make this more clear.

stijnmoreels: Discussed offline, we can keep on doing this adding of dependency ID's to the dependency tracking.

fgheysels: Just to be sure: we also have an overload which doesn't have the dependencyId; that overload generates an id himself ?
If I look at the code, that is not the case.

stijnmoreels: > Just to be sure: we also have an overload which doesn't have the dependencyId; that overload generates an id himself ?

No, we have an overload w/o passing a dependency ID but we can't generate this as it also has to be passed-along with the HTTP request, the Azure Service Bus message, .... or other 'request message' towards a dependent system.

fgheysels: Yes.  I get it.

fgheysels (approved):

PR #347, opened 2022-03-03
chore: use file paths io url routes

stijnmoreels:
Use relative file paths instead of URL routes for more stable linking of sub-pages.

Related to https://github.com/arcus-azure/arcus/issues/191

### DIFF
--- a/docs/preview/02-Features/sinks/azure-application-insights.md
+++ b/docs/preview/02-Features/sinks/azure-application-insights.md
@@ -153,7 +153,7 @@ var host = Host.CreateDefaultBuilder()
 
                     string instrumentationKey = context.Configuration["ApplicationInsights:InstrumentationKey"];
 
-                    if (!String.IsNullOrWhiteSpace(instrumentationKey))
+                    if (!string.IsNullOrWhiteSpace(instrumentationKey))
                     {
                         loggerConfiguration.WriteTo.AzureApplicationInsights(instrumentationKey, LogEventLevel.Information);
                     }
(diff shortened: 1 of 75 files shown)


docs/preview/02-Features/sinks/azure-application-insights.md
```diff
@@ -153,7 +153,7 @@ var host = Host.CreateDefaultBuilder()
 
                     string instrumentationKey = context.Configuration["ApplicationInsights:InstrumentationKey"];
 
-                    if (!String.IsNullOrWhiteSpace(instrumentationKey))
+                    if (!string.IsNullOrWhiteSpace(instrumentationKey))
```
stijnmoreels, line 156: Changed this to lower-case, bc it's also the way we do it in Arcus, and I think it's also recommended by Microsoft.

docs/preview/03-Guidance/use-with-dotnet-and-functions.md
```diff
@@ -54,16 +54,13 @@ namespace Arcus.Samples.AzureFunction
 
             builder.Services.AddLogging(loggingBuilder =>
             {
-                loggingBuilder.ClearProvidersExceptFunctionProviders();
```
stijnmoreels, line 57: Since we deprecated this method, let's also remove it in our docs.

fgheysels: Thought I did that, but must have forgotten it.

docs/preview/02-Features/sinks/azure-application-insights.md
fgheysels: Idk what is 'recommended'.  Anyway, there should be no difference as `String` refers to `System.String` and `string` is just an alias for it.
cfr `long` -> `System.Int64` , `int` -> `System.Int32` etc...

fgheysels (approved):

PR #350, opened 2022-03-15
chore: add CodeQL code analysis

stijnmoreels:
Add default CodeQL code analysis file.

Relates to https://github.com/arcus-azure/arcus/discussions/195

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,70 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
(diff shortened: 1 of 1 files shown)


fgheysels (commented):
Yes, let's do this.
I'm curious to see how the results of this code analysis are presented

stijnmoreels:
> Yes, let's do this. I'm curious to see how the results of this code analysis are presented.
> 
> I do see that this task is executed, but can we somehow easily see the results ?

When clicking on `details` in the 'Code scanning results` check.
https://github.com/arcus-azure/arcus.observability/pull/350/checks?check_run_id=5549335753

fgheysels:
> > Yes, let's do this. I'm curious to see how the results of this code analysis are presented.
> > I do see that this task is executed, but can we somehow easily see the results ?
> 
> When clicking on `details` in the 'Code scanning results` check. https://github.com/arcus-azure/arcus.observability/pull/350/checks?check_run_id=5549335753
> 
> It would maybe be nice to have it presented as a comment in the PR, but then again, it would mean a lot of clutter every time you commit to a PR and would be notified about it. When the scanning fails, it would stop the build, and I think that's enough trigger to look what's wrong. 👍

True.   I wonder though if there's an overview of the things that are being checked.

fgheysels (approved):

Issue #356, opened 2022-03-24
LogRequest is not logging the correct start-time


fgheysels:
This is related to https://github.com/arcus-azure/arcus.webapi/issues/310

When calling `LogRequest`, the timestamp at which the request was started, is not correctly logged.   (See the related issue for more detailed information).

This is because the `LogRequest` method does not contain a parameter that represents the `startDateTime`, and further up the chain, `DateTimeOffset.UtcNow` is used for determining the start-date of the request, which is obviously wrong, as the time at which you call this `LogRequest` method, is not necessarily at the time the request starts.

To fix this, we need to pass in the correct request-starttime to the `LogRequest` method.   There are different approaches for this:

- add yet another parameter to the `LogRequest` method which represents the start-time of the request.  Drawback of this approach is that the method signature becomes bloated.   Existing `LogRequest` overloads can call the newly introduced version and pass in `DateTimeOffset.UtcNow` as start-time.   However, these existing methods should be marked as obsolete imho.

- Instead of passing in a `Timespan` parameter to represent the duration, we could pass in a `DurationMeasurement` type.  We would then also need to extend the `DurationMeasurement` so that it contains a `StartedAt` or `StartDateTime` property which returns the start-timestamp of the measurement.   Drawback of this solution is that people need to pass in a `DurationMeasurement`, and that might not always be available.  (This can offcourse be solved by making sure that we can instantiate `DurationMeasurement` instances by passing in a DateTimeOffset and a Timespan when constructing it.  (Either via a constructor or via a factory method).

- We can pass - in the start-datetime as a property in the `context` parameter, but that is just implicit and not favorable.

stijnmoreels:
I think the draw-back of using `DurationMeasurement` is wrong here bc it already contains a `StartTime`/`Duration` so we can create overloads with that type and be done with it 🥳 .

PR #357, opened 2022-03-30
feat: use duration measurement when tracking http requests

stijnmoreels:
Use the `DurationMeasurement` when tracking HTTP requests.

Closes https://github.com/arcus-azure/arcus.observability/issues/356


src/Arcus.Observability.Telemetry.AspNetCore/Extensions/ILoggerExtensions.cs
```diff
@@ -44,6 +45,65 @@ public static void LogRequest(this ILogger logger, HttpRequest request, HttpResp
...
+        /// <exception cref="ArgumentOutOfRangeException">
+        ///     Thrown when the <paramref name="response"/>'s status code is outside the 0-999 range inclusively,
+        ///     or the <paramref name="duration"/> is a negative time range.
+        /// </exception>
+        public static void LogRequest(
+            this ILogger logger,
+            HttpRequest request,
+            HttpResponse response,
+            TimeSpan duration,
+            DateTimeOffset startTime,
```
fgheysels, line 94: Nitpicking:  it might be more logical to specify the startTime parameter before the duration parameter ?

stijnmoreels: Ah, yes, I based myself on how we track Azure Service Bus requests, but apparently, we do indeed place the start time in dependency tracking first.
Hmm, shame that we have missed it in Azure Service Bus request tracking 😢 . However, we recommend using the `DurationMeasurement` either way, so I guess it's not so bad.

Ok, let's change this! 👍

fgheysels (approved):
I think we're good to go!

stijnmoreels:
> I think we're good to go!

I was reworking some of the integration tests for a better/more stable approach. Then, with your blessing, we can merge this indeed 😉 .

stijnmoreels:
Voila! 🥳

PR #362, opened 2022-04-11
feat: add dependency id to custom dep tracking

stijnmoreels:
Adds dependency ID to custom dependency tracking.

Relates to #243


stijnmoreels:
Codefactor is complaining that our `ILogger` extensions class may be too complex. Probably due to the long list of extensions. Maybe we can make it partials based on type that is being logged?

fgheysels:
> Codefactor is complaining that our `ILogger` extensions class may be too complex. Probably due to the long list of extensions. Maybe we can make it partials based on type that is being logged?

Codefactor suggests some refactorings for this: https://www.codefactor.io/repository/github/arcus-azure/arcus.observability/pull/362

I don't think putting those extensions in separate partial classes will solve this.  I think Codefactor is more complaining about the [cyclomatic complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity) of the method implementation(s).

fgheysels (approved):

PR #366, opened 2022-04-25
chore: split ilogger extensions in partial classes

stijnmoreels:
Split the `ILogger` extensions in partial classes to minimize complexity and maximize maintanence.


fgheysels:
So, this solves the CodeFactor issue ?
I was wondering if splitting it in multiple files (partial classes) would solve this, as in the end, it is all compiled in the same class).  (See also my comment on another PR [here](https://github.com/arcus-azure/arcus.observability/pull/362#issuecomment-1112032629)).

fgheysels (approved):

stijnmoreels:
> So, this solves the CodeFactor issue ? I was wondering if splitting it in multiple files (partial classes) would solve this, as in the end, it is all compiled in the same class). (See also my comment on another PR [here](https://github.com/arcus-azure/arcus.observability/pull/362#issuecomment-1112032629)).

Yes, this solved that issue. It's more readable in any case, so I'm all for it 😄 .

PR #374, opened 2022-05-23
feat: add extension to rm microsoft appinsights logprovider

### LINKED ISSUE #364: Removing ClearProvidersExceptFunctionProviders causes the Application Insights OperationName to change
**Describe the bug**
Since the package Arcus.Observability.Telemetry.AzureFunctions has been marked as obsolete I have removed this package from my Azure Function. This meant I had to remove the `ClearProvidersExceptFunctionProviders` functionality from the startup. 

As a result the OperationName in Application Insights has changed from for example `GET /api/person/12345` to the FunctionName defined in `[FunctionName("MyFunction")]`.
Even specifying the OperationName to a custom value in the `LogRequest` results in the FunctionName and not the custom value. 
This means the overview of performance in Application Insights becomes unusable as it only shows the FunctionName instead of the API path.

**To Reproduce**
Remove `ClearProvidersExceptFunctionProviders` functionality from the startup.

**Expected behavior**
The OperationName stays the same.

**Additional context**
Latest versions of all the Arcus.Observability packages.

stijnmoreels:
Adds dedicated extension to remove Microsoft's `ApplicationInsightsLoggerProvider` to ensure that Arcus is in full control of telemetry.

Closes #364


src/Arcus.Observability.Tests.Integration/Serilog/Sinks/ApplicationInsights/AzureFunctionsDockerTests.cs
```diff
@@ -0,0 +1,59 @@
...
+                    await RetryAssertUntilTelemetryShouldBeAvailableAsync(async () =>
+                    {
+                        EventsResults<EventsRequestResult> results = await client.Events.GetRequestEventsAsync(ApplicationId, PastHalfHourTimeSpan);
+                        Assert.NotEmpty(results.Value);
+                        AssertX.Any(results.Value, result =>
+                        {
+                            Assert.Contains("order", result.Request.Url);
+                            Assert.Equal(200.ToString(), result.Request.ResultCode);
+                            Assert.True(Guid.TryParse(result.Request.Id, out Guid _));
+                            Assert.Equal(HttpMethod.Get.Method + " /api/order", result.Operation.Name);
```
stijnmoreels, line 52: Added a test here, Pim, so that we are sure that the operation name doesn't change again in the future. 😉

docs/preview/03-Guidance/use-with-dotnet-and-functions.md
```diff
@@ -12,8 +12,11 @@ We encourage you to follow the standard [Serilog instructions](https://github.co
 Some aspects we would like to highlight are:
 
 - Make sure to call [`UseSerilog`](https://www.nuget.org/packages/Serilog.AspNetCore) when creating a `IHostBuilder`
+- Remove the default Microsoft's `ApplicationInsightsLoggerProvider` via the `RemoveMicrosoftApplicationInsightsLoggerProvider` extension
 - Remove default for logging including its configuration in `appsettings.json` *(if applicable)*
```
fgheysels, line 16: What exactly is meant with this last bullet ? (remove default for logging ... ) ?

docs/preview/03-Guidance/use-with-dotnet-and-functions.md
```diff
@@ -12,8 +12,11 @@ We encourage you to follow the standard [Serilog instructions](https://github.co
...
 
+> We need to call `RemoveMicrosoftApplicationInsightsLoggerProvider` to remove Microsoft's `ApplicationInsightsLoggerProvider` because it would conflict with our own Serilog Application Insights sink. We can't garantee stable telemetry if Microsoft's logger provider is registered as this provider manipulates the telemetry before it get's send out to Application Insights. Removoving it ensure that Arcus is in full control of the send-out telemetry.
```
fgheysels, line 18:
```suggestion
> We need to call `RemoveMicrosoftApplicationInsightsLoggerProvider` to remove Microsoft's `ApplicationInsightsLoggerProvider` because it would conflict with our own Serilog Application Insights sink. We can't guarantee stable telemetry if Microsoft's logger provider is registered as this provider manipulates the telemetry before it get's send out to Application Insights. Removing it ensure that Arcus is in full control of the send-out telemetry.
```

src/Arcus.Observability.Telemetry.AzureFunctions/Extensions/ILoggerBuilderExtensions.cs
```diff
@@ -37,5 +38,32 @@ public static ILoggingBuilder ClearProvidersExceptFunctionProviders(this ILoggin
...
+        public static ILoggingBuilder RemoveMicrosoftApplicationInsightsLoggerProvider(this ILoggingBuilder builder)
+        {
+            Guard.NotNull(builder, nameof(builder), "Requires an instance of the logging builder to remove Microsoft's 'ApplicationInsightsLoggerProvider'");
+
+            ServiceDescriptor descriptor = 
+                builder.Services.FirstOrDefault(service => service.ImplementationType?.Name == "ApplicationInsightsLoggerProvider");
+            
+            if (descriptor is null)
+            {
+                throw new InvalidOperationException(
```
fgheysels, line 59: I wonder why we throw an exception if we cannot find something that we want to remove .... 

The only thing I can think of, is to make sure that the code fails should MSFT decide to rename this servcie, which makes it valid to throw then ....

src/Arcus.Observability.Tests.Integration/Serilog/Sinks/ApplicationInsights/AzureFunctionsDockerTests.cs
```diff
@@ -0,0 +1,59 @@
...
+                using (ApplicationInsightsDataClient client = CreateApplicationInsightsClient())
+                {
+                    await RetryAssertUntilTelemetryShouldBeAvailableAsync(async () =>
+                    {
+                        EventsResults<EventsRequestResult> results = await client.Events.GetRequestEventsAsync(ApplicationId, PastHalfHourTimeSpan);
+                        Assert.NotEmpty(results.Value);
+                        AssertX.Any(results.Value, result =>
+                        {
+                            Assert.Contains("order", result.Request.Url);
+                            Assert.Equal(200.ToString(), result.Request.ResultCode);
```
fgheysels, line 50: Why not just write this as : 
```suggestion
                            Assert.Equal("200", result.Request.ResultCode);
```
?

docs/preview/03-Guidance/use-with-dotnet-and-functions.md
stijnmoreels: Was also thinking about that. I think it's the `"applicationInsights"` parts in the `appsettings.json`, but I can't quite remember why it was here.

src/Arcus.Observability.Telemetry.AzureFunctions/Extensions/ILoggerBuilderExtensions.cs
stijnmoreels: Yes, it's for the reason you mentioned. If we don't do anything in this case, the failure may be 'swallowed' and we would probably have obscure failures (like faulty operation names mentioned in #364 ).

src/Arcus.Observability.Tests.Integration/Serilog/Sinks/ApplicationInsights/AzureFunctionsDockerTests.cs
stijnmoreels: Yeah, is same for me. Integer for the code, as it's a number, and not text. But it's ok your suggestion, too.

src/Arcus.Observability.Tests.Integration/Serilog/Sinks/ApplicationInsights/AzureFunctionsDockerTests.cs
pim-simons: Great stuff 👍🏻

src/Arcus.Observability.Telemetry.AzureFunctions/Extensions/ILoggerBuilderExtensions.cs
fgheysels: On the other hand, you could argue that this is something that we should detect with a test.

How is this Application Insights provider being added to the DI container ?  Is it automatically there ?  Is it added via a call to the `AddApplicationInsights` method ?  What if we have a situation where the App Insights provider is actually not there ?  Do we need to fail on that ?

stijnmoreels: Yeah, it's by default there, even without any app settings or startup content. If we start investigating in out-of-process Azure Functions, we can come back on this. Maybe in those cases, it would not be needed.

fgheysels: I'm just reluctant to throw here if this situation arises.
It means that your complete Function won't be able to work, just because your logging is maybe not working optimally.  I think it might be better to log a Warning instead.

stijnmoreels: Yeah, I don't think we can log here already bc the application is not yet build up. Maybe we can log a DEBUG trace, though.

stijnmoreels: The alternative, is that the logging is corrupt if we don't throw.

fgheysels: A Debug Trace will not be logged when your loglevel is set to Information or Warning , which is most likely the case for production API's.   Therefore, I'd log a warning or error.

The alternative is corrupt logging, but a failing API and a support call for such an issue is also not something you want.  On the other hand, these are things that should pop-up during development / testing.

stijnmoreels: Yeah, but like I said: we cannot log an error/warning here the 'normal' way because the logging is at this place still being constructed. We should look for other ways to inform the consumer. A DEBUG trace is an option, not sure what's left.

fgheysels: Had a chat with @gverstraete .  Conclusion was that, if the provider is not there, the API will fail to start.  Since it is at startup time, it should be obvious that something is wrong.  Chances that this occurs on a restart of a working API are rather scarce.   
Hence, we'll just throw.

docs/preview/03-Guidance/use-with-dotnet-and-functions.md
fgheysels: Should we remove this bullet, as it is only causing confusion ?
Or should we try to clarify it a bit.   The `Logging` [section ](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-6.0#configure-logging)in appsettings.json is indeed not used when we use Serilog, as it uses another section.

Maybe we should modify this piece to make it more clear.  Something like:
```suggestion
- Remove the `Logging` section from the `appsettings.json` *(if applicable)* as this is not used by Serilog
```

fgheysels (approved):

src/Arcus.Observability.Telemetry.AzureFunctions/Extensions/ILoggerBuilderExtensions.cs
stijnmoreels: Ok, great! Thx also @gverstraete to look into this.

docs/preview/03-Guidance/use-with-dotnet-and-functions.md
stijnmoreels: Yeah, ok! It's the `Serilog` section in the `appsettings.json` file that Serilog can use with the `ReadFrom.Configuration(...)`, right?

src/Arcus.Observability.Tests.Integration/Serilog/Sinks/ApplicationInsights/AzureFunctionsDockerTests.cs
fgheysels: I was just wondering , as "200" seems more straightforward then `200.ToString` :)

docs/preview/03-Guidance/use-with-dotnet-and-functions.md
fgheysels: Yes indeed.

PR #376, opened 2022-05-24
feat: add dependency id to sql dep tracking

### LINKED ISSUE #286: Improve the LogSqlDependency method signature
The `LogSqlDependency` extension method on `ILogger` has a parameter which is called `tableName`.

This is a bit shortcoming as, when you want to log a dependency on SQL server, you execute (in most cases) a query that touches multiple tables.
This `tableName` parameter is used to build up the `dependencyName`:

image

It makes no sense to pass in a single tablename if you perform a SQL query that touches multiple tables, so it would be better if we:

- rename the `tableName` parameter to something more generic:  `sqlCommand`  or `sqlCommandInfo`
- Since that parameter can contain a lot more information, it should not be used to build up the `dependencyName`.  Only the database-name should be in the dependency name.
- The `sqlCommand` / `sqlCommandInfo` parameter can be logged as a contextual item / dimension
- The `sqlCommand`/ `sqlCommandInfo`  parameter should not be considered as mandatory, so we should not guard it.

See also this dicussion: https://github.com/arcus-azure/arcus.observability/discussions/265

stijnmoreels:
Adds dependency ID to SQL dependency tracking.

Closes #286
Relates to #243

### DIFF
--- a/src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerSqlDependencyExtensions.cs
+++ b/src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerSqlDependencyExtensions.cs
@@ -80,6 +80,66 @@ namespace Microsoft.Extensions.Logging
         }
 
         /// <summary>
+        /// Logs a SQL dependency.
+        /// </summary>
+        /// <param name="logger">The logger to track the telemetry.</param>
+        /// <param name="serverName">The name of server hosting the database.</param>
+        /// <param name="databaseName">The name of database.</param>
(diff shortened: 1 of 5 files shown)


stijnmoreels:
Adding this overload could cause some ambiguous overloads if both the Core and Sql package are installed without argument naming.

fgheysels (approved):

stijnmoreels:
> Adding this overload could cause some ambiguous overloads if both the Core and Sql package are installed without argument naming.

@fgheysels so you think this is not a problem we should worry about?

PR #377, opened 2022-05-25
docs: update w better metric value

### LINKED ISSUE #281: Improve LogMetric documentation
It would be nice if the sample that is used for the `LogMetrics` method actually used something that looks more like a metric.
The metric is now called 'Invoice Received', and the value is 133.37.   This doesn't really look like a metric to me.   I would interpret it as a 'an invoice with an amount of 133.37 is received'.

However, `LogMetric` should really log ... metrics.   For instance the amount of invoices received or processed.   Or the time it took to process an invoice, ....

[This](https://observability.arcus-azure.net/features/writing-different-telemetry-types/#metrics) is a link to the documentation in question.

stijnmoreels:
Improves the metric logging feature docs with a better description of a metric.

Closes #281


fgheysels (approved):
Small improvement, but absolutely makes sense!

PR #386, opened 2022-06-03
feat: add upstream service correlation options

stijnmoreels:
This will also make things a better streamlined as the operation parent ID was already in the Serilog correlation enrichers but not in the correlation model itself.
Currently, only the HTTP correlation properties has this upstream service, but with this in place, we wouldn't need that anymore.

Related to https://github.com/arcus-azure/arcus.templates/issues/575


stijnmoreels:
Will maybe need another name, too, as it's the same name as the HTTP correlation options, now. I'll also add some unit tests here.
- [ ] Propose new name
- [ ] Add unit tests.

PR #411, opened 2022-06-17
feat: diff request & other correlation assignments

### LINKED ISSUE #410: Differentiate between request and other types of telemetry when assigning correlation
**Is your feature related to a problem? Please describe.**
Currently, we don't differentiate between request and other types of telemetry when we assign the correlation information before we track the telemetry in Application Insights. The leads to a problem of parent linking as the dependencies have the same parent as the request, while the dependencies (and other types) should have as parent the incoming request.

**What solution do you propose?**
We should differentiate between request telemetry and other types of telemetry in the `OperationContextConverter` in the `Arcus.Observability.Telemetry.Serilog.Sinks.ApplicationInsights` project.

When assigning correlation to request telemetry, these properties should be affected:
- `Id` = `Correlation.OperationId` or generated from request tracking options if not defined
- `Context.Operation.Id` = `Correlation.TransactionId`
- `Context.Operation.ParentId` = `Correlation.OperationParentId`

When assigning correlation to other types of telemetry, these properties should be affected:
- `Context.Operation.Id` = `Correaltion.TransactionId`
- `Context.Operation.ParentId` = `Correlation.OperationId`

stijnmoreels:
Differentiate between request and other types of telemetry tracking.

Closes #410

### DIFF
--- a/src/Arcus.Observability.Telemetry.Serilog.Sinks.ApplicationInsights/Converters/RequestTelemetryConverter.cs
+++ b/src/Arcus.Observability.Telemetry.Serilog.Sinks.ApplicationInsights/Converters/RequestTelemetryConverter.cs
@@ -19,6 +19,7 @@ namespace Arcus.Observability.Telemetry.Serilog.Sinks.ApplicationInsights.Conver
         /// <summary>
         /// Initializes a new instance of the <see cref="RequestTelemetryConverter" /> class.
         /// </summary>
+        [Obsolete("Use the constructor overload with the Application Insights options instead")]
         public RequestTelemetryConverter()
             : this(new ApplicationInsightsSinkRequestOptions())
         {
@@ -29,12 +30,22 @@ namespace Arcus.Observability.Telemetry.Serilog.Sinks.ApplicationInsights.Conver
         /// </summary>
         /// <param name="options">The user-defined configuration options to tracking requests.</param>
         /// <exception cref="ArgumentNullException">Thrown when the <paramref name="options" /> is <c>null</c>.</exception>
+        [Obsolete("Use the constructor overload with the Application Insights options instead")]
         public RequestTelemetryConverter(ApplicationInsightsSinkRequestOptions options)
         {
             Guard.NotNull(options, nameof(options), "Requires a set of user-configurable options to influence the behavior of how requests are tracked");
             _options = options;
         }
-        
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="RequestTelemetryConverter" /> class.
+        /// </summary>
+        /// <param name="options">The user-defined configuration options to influence the behavior of the Application Insights Serilog sink.</param>
+        /// <exception cref="ArgumentNullException">Thrown when the <paramref name="options"/> is <c>null</c>.</exception>
+        public RequestTelemetryConverter(ApplicationInsightsSinkOptions options) : base(options)
+        {
+        }
+
         /// <summary>
         ///     Creates a telemetry entry for a given log event
         /// </summary>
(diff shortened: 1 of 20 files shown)


src/Arcus.Observability.Telemetry.Serilog.Sinks.ApplicationInsights/Converters/RequestTelemetryConverter.cs
```diff
@@ -67,12 +76,16 @@ protected override RequestTelemetry CreateTelemetryEntry(LogEvent logEvent, IFor
 
             var requestTelemetry = new RequestTelemetry(sourceName, requestTime, requestDuration, responseStatusCode, isSuccessfulRequest)
             {
-                Id = id,
                 Url = url,
                 Source = source,
                 Context = { Operation = { Name = requestOperationName } }
             };
 
+            if (_options != null)
```
fgheysels, line 84: Can `_options` ever be null ?

src/Arcus.Observability.Tests.Integration/Serilog/Sinks/ApplicationInsights/EventTests.cs
```diff
@@ -161,8 +161,8 @@ public async Task LogEventWithCorrelationInfo_SinksToApplicationInsights_Results
                         Assert.Equal(operationId, actualOperationId);
                         Assert.Equal(transactionId, actualTransactionId);
                         Assert.Equal(operationParentId, actualOperationParentId);
-                        Assert.Equal(operationId, trace.Operation.Id);
-                        Assert.Equal(operationParentId, trace.Operation.ParentId);
+                        Assert.Equal(transactionId, trace.Operation.Id);
+                        Assert.Equal(operationId, trace.Operation.ParentId);
```
fgheysels, line 165: I might be missing some context, but I think this is strange ?

src/Arcus.Observability.Telemetry.Serilog.Sinks.ApplicationInsights/Converters/RequestTelemetryConverter.cs
stijnmoreels: Yes, bc this implementation makes sure we support both the 'old' way of injecting the request-specific options, and the 'new' way of options passed-along via the base class.

src/Arcus.Observability.Tests.Integration/Serilog/Sinks/ApplicationInsights/EventTests.cs
stijnmoreels: This is the conclusion from the service-to-service PR we had last week. The non-requests will need some different telemetry correlation to be fully linked to the incoming requests.

Also, Application Insights' `Operation ID` = `Transaction ID` from Arcus.

src/Arcus.Observability.Telemetry.Serilog.Sinks.ApplicationInsights/Converters/RequestTelemetryConverter.cs
stijnmoreels: The ID is now set in the `OperationContextConverter` where the correlation assignment is centralized. There, the operation ID is being used as the ID of the request (like proposed in the service-to-service poc).

src/Arcus.Observability.Tests.Integration/Serilog/Sinks/ApplicationInsights/EventTests.cs
stijnmoreels: 😄 I can only imagine what that would be like, constantly switching. You're doing a fine job! 👍

fgheysels (approved):
I think this is ok :)

PR #425, opened 2022-07-01
feat: add configurable serilog sink correlation options

### LINKED ISSUE #409: Application Insights sink doesn't take into account custom correlation property names
**Describe the bug**
When the Application Insights sink extracts the properties from the Serilog log event properties and assign it to the Application Insights telemetry model, it uses the default `ContextProperties.Correlation.xxx` values, while the `CorrelationInfoEnricher` can use custom Serilog property names to assign the correlation. 

**To Reproduce**

**Expected behavior**
We should make sure that the Application Insights sink can be configurable with custom correlation properties too, so that the consumer can use custom property names and the Application Insights sink can extract those correctly.

**Additional context**
 - v2.4

stijnmoreels:
Make correlation properties configurable in Azure Application Insights Serilog sink.

Closes #409


docs/preview/02-Features/sinks/azure-application-insights.md
```diff
@@ -112,6 +128,39 @@ With this configuration, the custom dimension name for the public properties of
...
+        // Sets a custom property name where the operation parent ID should be added (default: 'OperationParentId').
+        options.Correlation.OperationParentIdPropertyName = operationParentIdPropertyName;
+    });
+```
+
+For more information on the way how Arcus handles correlation, see [this dedicated page](../correlation.md).
+
 ## FAQ
 
 ### Q: Why is it mandatory to provide a instrumentation key?
```
fgheysels, line 166: This is no longer correct, as we also check if an instrumentationkey (or connectionstring) is provided.
At least, we check this in templates if I'm not mistaken.
The question might be a bit misleading.   Maybe it should be:

Why do I have to configure an instrumentationkey (connectionstring) when using the Application Insights sink

src/Arcus.Observability.Telemetry.Serilog.Sinks.ApplicationInsights/Configuration/ApplicationInsightsSinkCorrelationOptions.cs
```diff
@@ -0,0 +1,58 @@
...
+    /// User-defined configuration options to influence the behavior of the Azure Application Insights Serilog sink while assigning telemetry correlation.
+    /// </summary>
+    public class ApplicationInsightsSinkCorrelationOptions
+    {
+        private string _operationIdPropertyName = ContextProperties.Correlation.OperationId,
+                       _transactionIdPropertyName = ContextProperties.Correlation.TransactionId,
+                       _operationParentIdPropertyName = ContextProperties.Correlation.OperationParentId;
+
+        /// <summary>
+        /// Gets or sets the Serilog application property name where the correlation operation ID is set.
```
fgheysels, line 17: Should we specifiy 'correlation' here ?
Isn't it sufficient and more clear to do this:

```suggestion
        /// Gets or sets the Serilog application property name where the operation ID is set.
```

src/Arcus.Observability.Telemetry.Serilog.Sinks.ApplicationInsights/Configuration/ApplicationInsightsSinkCorrelationOptions.cs
```diff
@@ -0,0 +1,58 @@
...
+            get => _operationIdPropertyName;
+            set
+            {
+                Guard.NotNullOrWhitespace(value, nameof(value), "Requires a non-blank Serilog application property name to locate the correlation operation ID");
+                _operationIdPropertyName = value;
+            }
+        }
+
+        /// <summary>
+        /// Gets or sets the Serilog application property name where the correlation transaction ID is set.
```
fgheysels, line 31:
```suggestion
        /// Gets or sets the Serilog application property name where the transaction ID is set.
```

src/Arcus.Observability.Telemetry.Serilog.Sinks.ApplicationInsights/Configuration/ApplicationInsightsSinkCorrelationOptions.cs
fgheysels: Do we need to explain what the OperationID is ?

fgheysels (approved):

src/Arcus.Observability.Telemetry.Serilog.Sinks.ApplicationInsights/Configuration/ApplicationInsightsSinkCorrelationOptions.cs
stijnmoreels: We could, but we should do this for all mentions of the correlation ID's. Maybe we could provide a link to the documentation site that does the explanation better than trying to do the same in code docs that nobody (?) reads 😄

stijnmoreels: Will be added here: https://github.com/arcus-azure/arcus.observability/issues/431

docs/preview/02-Features/sinks/azure-application-insights.md
stijnmoreels: Done in https://github.com/arcus-azure/arcus.observability/pull/425/commits/aa110d02af14401a143f7b2f8d6740f36e69fba4

PR #426, opened 2022-07-01
docs: place guides above features

### LINKED ISSUE #422: doc: modify the order of the menu
It would be nice if we could modify the order of the topic-menu, and make sure that the 'Guidance' section is placed above the 'Features' section:

image

stijnmoreels:
Places the user guides above the features tab in the sidebar for both the preview and currently most recent docs (v2.5)

Closes #422


fgheysels (approved):
Thanks, I think this will be a great improvement

PR #427, opened 2022-07-01
fix: use correct deprecation az servicebus messages

### LINKED ISSUE #417: Tracking Azure Service Bus dependencies doesn't always have correct deprecated message
Topic and queues are mismatched when adding deprecated messages when tracking Azure Service Bus dependencies.

Make sure that the correct entity type is used in the deprecation messages.

stijnmoreels:
Fix the deprecation messages of the Azure Service Bus queue/topic/general dependency tracking extensions.

Closes #417


stijnmoreels:
A special thx to @pim-simons for noticing this 😉 .

fgheysels (approved):

Issue #407, opened 2022-06-13
Remove deprecated `LogTableStorageDependency` extensions


stijnmoreels:
**Is your feature request related to a problem? Please describe.**
Previously, we used the `DependencyMeasurement` to track dependencies but we have changed that to the `DurationMeasurement`.

**Describe the solution you'd like**
Remove the `LogTableStorageDependency` extension in the `Arcus.Observability.Telemetry.Core` project that still uses the `DependencyMeasurement`.

vtermaat:
picking this up

PR #434, opened 2022-07-08
feat: Remove deprecated LogTableStorageDependency extensions (🚧)

vincenttermaat:
Remove deprecated LogTableStorageDependency extensions
Closes https://github.com/arcus-azure/arcus.observability/issues/407


stijnmoreels (approved):
Thx Vincent for the many contributions! ❤️ Help us a lot.

vtermaat:
> > Thx Vincent for the many contributions! ❤️ Help us a lot.
> 
> I can only confirm :)
> 
> I'm just a bit hesitant on this.... If I'm not mistaken, we've only had one prior release where those methods were marked as obsolete. Aren't we a bit to eager to remove them already ? What about people that are upgrading Arcus and skipping one (or more) versions ?
> 
> OTOH, this is something that is going to be included in the next major, so breaking changes are allowed. What are your opinions on this @vtermaat @stijnmoreels ?

What are the conventions for this?
I mean. Whenever I upgrade from 2.x to 3.x I expect to do some rework. Most important for me would be that there are no bugs in 2.x that are causing issues for me. So I can plan for the upgrade and don't need to do rework in a rush.

stijnmoreels:
> I'm just a bit hesitant on this.... If I'm not mistaken, we've only had one prior release where those methods were marked as obsolete. Aren't we a bit to eager to remove them already ? What about people that are upgrading Arcus and skipping one (or more) versions ?
> 
> OTOH, this is something that is going to be included in the next major, so breaking changes are allowed. What are your opinions on this @vtermaat @stijnmoreels ?

It's true that we only did one version with the deprecated stuff. There is already a lot of stuf fthat's deprecated already, but we could however indeed hold of and first release v2.6 instead of v3 bc the main major changes for the next major release would be removing functionality and not adding. But again, it's a matter of preference. If we decide to do the major we will get rid of many deprecations, but if we decide to radically change some functionality, we would already be at v4 in no time.

PR #447, opened 2022-08-01
feat: track custom request

### LINKED ISSUE #438: Adds a custom request telemetry extension
When tracking requests in Application Insights, we can only choose from HTTP or Azure Service Bus (and in the future EventHubs), but we don't allow custom requests like we do custom dependencies.

Add `ILogger` extension called `LogCustomRequest` that takes in the necessary information and can be used for a wide range of systems.

stijnmoreels:
Adds extensions to track custom requests (like we can track custom dependencies) to let users make use of Observability outside provided systems we provide request tracking (API, Service Bus, EventHubs).

Closes #438

### DIFF
--- a/src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerEventHubsRequestExtensions.cs
+++ b/src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerEventHubsRequestExtensions.cs
@@ -37,7 +37,7 @@ namespace Microsoft.Extensions.Logging
             Guard.NotNullOrWhitespace(eventHubsName, nameof(eventHubsName), "Requires an Azure EventHubs name to track the request");
             Guard.NotNull(measurement, nameof(measurement), "Requires an instance to measure the Azure EventHubs request process latency duration");
 
-            LogEventHubsRequest(logger, eventHubsNamespace, eventHubsName, isSuccessful, measurement.Elapsed, measurement.StartTime, context);
+            LogEventHubsRequest(logger, eventHubsNamespace, eventHubsName, isSuccessful, measurement.StartTime, measurement.Elapsed, context);
         }
 
         /// <summary>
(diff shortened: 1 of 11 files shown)


src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerEventHubsRequestExtensions.cs
```diff
@@ -112,22 +113,23 @@ public static partial class ILoggerExtensions
...
         public static void LogEventHubsRequest(
             this ILogger logger,
             string eventHubsNamespace,
             string consumerGroup,
             string eventHubsName,
             string operationName,
             bool isSuccessful,
-            TimeSpan duration,
             DateTimeOffset startTime,
+            TimeSpan duration,
```
stijnmoreels, line 132: Updated the signature to be in line with the other extensions.
We can still do this as it's not yet released.

src/Arcus.Observability.Telemetry.Core/Logging/RequestSourceSystem.cs
```diff
@@ -18,6 +18,11 @@ public enum RequestSourceSystem
         /// <summary>
         /// Specifies that the request-source is an Azure EventHubs.
         /// </summary>
-        AzureEventHubs = 4
+        AzureEventHubs = 4,
+
+        /// <summary>
+        /// Specifies that the request-source is a custom system.
+        /// </summary>
+        Custom = 8
```
fgheysels, line 26: Is that backing integer used somewhere ?  Should we use a higher number ?

fgheysels (approved):

src/Arcus.Observability.Telemetry.Core/Logging/RequestSourceSystem.cs
stijnmoreels: It's mostly to be strict in using switches. If we add more types, unexpected scenario's could happen when interacting with the enum in a different way. It's basically a failsafe to make sure we know what value is behind each value.

fgheysels: Ok, that's why I was asking if we should use a higher number for 'custom'.  (Or is it not necessary to have the highest backing intger assigned to 'custom').

fgheysels: But, the enum is not a Flag - enum, so I don't see how you could apply bitwise operations to them.

fgheysels (approved):

src/Arcus.Observability.Telemetry.Core/Logging/RequestSourceSystem.cs
stijnmoreels: You could still use bitwise operations, it's just that if the number overlaps on a future number, we could have unexpected results.

stijnmoreels: > Ok, that's why I was asking if we should use a higher number for 'custom'. (Or is it not necessary to have the highest backing intger assigned to 'custom').

We could provide a higher number, that's true, to maybe make sure that it's cleaner, but on the other hand, we don't use these integers. Only as a failsafe.

PR #459, opened 2022-08-26
(❗) fix: null-check on scalar value

stijnmoreels:
Reason why the Azure Functions didn't always tracked the request, was that the operation parent ID was `null` in the initial call and this `null` wasn't checked when parsing the Serilog values to typed variants.


fgheysels (approved):
Nice find!

PR #460, opened 2022-08-29
(❗) fix: add operation name to sql dependency tracking

### LINKED ISSUE #419: Make `dependencyName` in SQL dependency tracking more unique
In a previous release, we made sure that we could pass-in a pseudo SQL command because it is quite common to interact with multiple SQL tables in a realistic application (JOIN, multiple SELECT...).
Although, this change cased a more general `dependencyName` for SQL dependency tracking, which messed up the performance overview as all dependency tracking is now under the same `databaseName`.

We should make sure that we again make the SQL dependency tracking more unique. We could do this by adding (for example) the pseudo SQL command to the `dependencyName`.

`<databaseName>/<sqlCommand`
`Orders/SELECT OrderId FROM Orders`

Discovered by @pim-simons .

stijnmoreels:
Make the dependency name unique by adding a functional description of the operation.

Closes #419

### DIFF
--- a/src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerSqlDependencyExtensions.cs
+++ b/src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerSqlDependencyExtensions.cs
@@ -59,7 +59,7 @@ namespace Microsoft.Extensions.Logging
         /// <exception cref="ArgumentException">
         ///     Thrown when the <paramref name="serverName"/>, <paramref name="databaseName"/>, <paramref name="tableName"/>, or <paramref name="operationName"/> is blank.
         /// </exception>
-        [Obsolete("Use the " + nameof(LogSqlDependency) + " with a pseudo SQL command instead of specifying the table and operation name")]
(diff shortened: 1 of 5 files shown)


src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerSqlDependencyExtensions.cs
```diff
@@ -271,5 +275,88 @@ public static partial class ILoggerExtensions
...
+            Guard.NotNullOrWhitespace(serverName, nameof(serverName), "Requires a non-blank SQL server name to track a SQL dependency");
+            Guard.NotNullOrWhitespace(databaseName, nameof(databaseName), "Requires a non-blank SQL database name to track a SQL dependency");
+            Guard.NotLessThan(duration, TimeSpan.Zero, nameof(duration), "Requires a positive time duration of the SQL dependency operation");
+
+            context = context ?? new Dictionary<string, object>();
+
+            logger.LogWarning(MessageFormats.SqlDependencyFormat, new DependencyLogEntry(
+                dependencyType: "Sql",
+                targetName: serverName,
+                dependencyName: databaseName + "/" + operationName,
```
stijnmoreels, line 352: Most important change here.

fgheysels (approved):

PR #462, opened 2022-08-29
(❗) chore: rework correlation info options

### LINKED ISSUE #418: Rework general correlation options
The current correlation options is now mostly used by Web API but is not really used by Messaging. The problem is that changes in Web API should go via observability, plus that HTTP-related concepts creep in the general correlation options, while those concepts aren't available in other contexts like Messaging.

We already have introduced a HTTP-specific correlation options model to Web API, so we could more easily move away from the general options, now, we should make sure that we remove HTTP-specific concepts from the observability ones.

stijnmoreels:
Remove any implementation-specific options or registrations so that the correlation is purely an abstration, which is what it's supposed to be.

Closes #418


src/Arcus.Observability.Correlation/IServiceCollectionExtensions.cs
```diff
@@ -11,29 +11,60 @@ namespace Microsoft.Extensions.DependencyInjection
...
             return AddCorrelation(services, new DefaultCorrelationInfoAccessor(), configureOptions);
         }
 
+        /// <summary>
+        /// Adds operation and transaction correlation to the application using the <see cref="DefaultCorrelationInfoAccessor{TCorrelationInfo}"/>
+        /// </summary>
+        /// <typeparam name="TCorrelationInfo">The type of the <see cref="CorrelationInfo"/> model.</typeparam>
+        /// <param name="services">The services collection containing the dependency injection services.</param>
+        /// <exception cref="ArgumentNullException">Thrown when the <paramref name="services"/> is <c>null</c>.</exception>
+        public static IServiceCollection AddCorrelation<TCorrelationInfo>(
```
fgheysels, line 47: I don't immediately see how this must be used now ?
You need to create a type which inherits from CorrelationInfo, or do we have some predefined types ?

stijnmoreels: Yes, this is used when you want to use a custom correlation model. This extension is used, for example, in the message pump where we use `MessageCorrelationInfo`.

fgheysels (approved):

src/Arcus.Observability.Correlation/IServiceCollectionExtensions.cs
fgheysels: ok, thx

Issue #470, opened 2022-10-12
Extract application/component name into a service so that both Arcus' Serilog as well as Microsoft `TelemetryClient` can benefit from this


stijnmoreels:
**Is your feature request related to a problem? Please describe.**
When combining Arcus Serilog Application Insights sink and Microsoft's `TelemetryClient`, the application/component name is different or duplicated because Arcus works with enrichers to assign a component name and the `TelemetryClient` requires a registered `ITelemetryInitializer`.

**Describe the solution you'd like**
Extract the application/component name into a service so that both Arcus' Serilog as well as the Microsoft `TelemetryClient` can benefit from this and we only have to configure this once.

PR #472, opened 2022-10-12
feat: add re-usable application name

stijnmoreels:
Extract application/component name so that both Arcus and Microsoft functionality can benefit from this.

Closes https://github.com/arcus-azure/arcus.observability/issues/470

### DIFF
--- a/src/Arcus.Observability.Telemetry.Serilog.Enrichers/Extensions/LoggerEnrichmentConfigurationExtensions.cs
+++ b/src/Arcus.Observability.Telemetry.Serilog.Enrichers/Extensions/LoggerEnrichmentConfigurationExtensions.cs
@@ -93,6 +93,36 @@ namespace Serilog
         }
 
         /// <summary>
+        /// Adds the <see cref="ApplicationEnricher"/> to the logger enrichment configuration which adds the given application's name,
+        /// where the name is retrieved from a registered <see cref="IAppName"/> in the <paramref name="serviceProvider"/>.
+        /// </summary>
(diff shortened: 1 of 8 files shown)


src/Arcus.Observability.Telemetry.Serilog.Enrichers/Extensions/LoggerEnrichmentConfigurationExtensions.cs
```diff
@@ -92,6 +92,36 @@ public static LoggerConfiguration WithVersion(this LoggerEnrichmentConfiguration
             return enrichmentConfiguration.With(new ApplicationEnricher(componentName, propertyName));
         }
 
+        /// <summary>
```
fgheysels, line 95: Should we deprecate the other WithComponentName method ?
Or can we modify the implementation of the `WithComponentName` method that takes a string so that an `AppName` service is registered etc ?  That will be difficult I guess, since you do not have your service-collection there.

stijnmoreels: Yes, that's right, it will be difficult as the logging is set up after the `AppName` is registered. Didn't deprecate this bc there are many valid configurations. You could also pass it here and don't use W3C, or pass in the name two times, or use a slightly different name for Microsoft technology. Also, if there's no need for automatic dependency tracking, you could also pass it here directly.

fgheysels (approved):

PR #485, opened 2022-11-14
chore: use new microsoft sql package

stijnmoreels:
Use new Microsoft SQL package in Telemetry SQL project.

### DIFF
--- a/src/Arcus.Observability.Telemetry.Sql/Arcus.Observability.Telemetry.Sql.csproj
+++ b/src/Arcus.Observability.Telemetry.Sql/Arcus.Observability.Telemetry.Sql.csproj
@@ -24,7 +24,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="System.Data.SqlClient" Version="4.8.1" />
+    <PackageReference Include="Microsoft.Data.SqlClient" Version="5.0.1" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/Arcus.Observability.Telemetry.Sql/Extensions/ILoggerExtensions.cs
+++ b/src/Arcus.Observability.Telemetry.Sql/Extensions/ILoggerExtensions.cs
@@ -1,8 +1,8 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Data.SqlClient;
 using Arcus.Observability.Telemetry.Core;
 using GuardNet;
+using Microsoft.Data.SqlClient;
 
 // ReSharper disable once CheckNamespace
 namespace Microsoft.Extensions.Logging


src/Arcus.Observability.Telemetry.Sql/Arcus.Observability.Telemetry.Sql.csproj
```diff
@@ -24,7 +24,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="System.Data.SqlClient" Version="4.8.1" />
+    <PackageReference Include="Microsoft.Data.SqlClient" Version="5.0.1" />
```
fgheysels, line 27: I agree that we should use the newer version of this, but I also wonder what kind of impact this will have on users of Arcus.Observability.
We actually only have this dependency because we use the `SqlConnectionStringBuilder` class in one of the overloads of the `LogSqlDependency` that takes a connection-string as parameter.
Ppl that are using Arcus.Observability and are using System.Data.SqlClient themselves will implicitly import this dependency and might suffer issues because it might not be clear if they're now using System* or Microsoft.Data.SqlClient classes.

I would be in favor if we could get rid of this dependency in Arcus.Observability all together :)

stijnmoreels: > I would be in favor if we could get rid of this dependency in Arcus.Observability all together :)

Than we would not have the SQL-specific overloads in the `.Sql` sub-package.

Maybe it is also the responsibility of the consumer to know where their dependencies are coming from and not solely rely on implicit ones.

fgheysels: OTH, I think it is a bit bloated to have a dependency on this package, just to be able to parse connection-strings.
We just want toretrieve some information from the connection-string.   I wonder if we cannot do that ourselves.

The connection-string is nothing more than a string that contains multiple key/value pairs, separated by a semi-colon.

stijnmoreels: Yeah, but if you look at the implementation of the SQL connection string builder, it is rather large and has a lot of options and aliases. If we would not use the already built-in connection string builder, we have to create a rather lot of extensive testing to make sure that we can include all those different types, and we still be able to retrieve the data source and initial catalog; all the while when there is already such an implementation built-in. Not sure if it would be worth the effort if the benefit would be to remove this dependency (which is only added if you explicitly use the SQL-specific package, not the Core package which also has an SQL extension).

fgheysels: IDK by heart what we are retrieving from the connectionstring, I believe it is just the servername ?

stijnmoreels: Will have to investigate how much work we would have to do for this to be stable.
It is the data source and initial catalog that we retrieve from the connection string. 
Currently, this PR was to make sure that step away from the older SQL package (which already has a fix release via dependabot in another PR).
So, we could either close this PR, investigate how much work it would be to implement it ourselves, and merge [this](https://github.com/arcus-azure/arcus.observability/pull/484) PR to fix the older SQL package.
Our we could merge this PR, and possible investigate how much work it would be to implement it ourselves.

fgheysels: Let's merge and investigate.

fgheysels (approved):

PR #490, opened 2022-11-30
chore: sql property parsing io sql package

### LINKED ISSUE #488: Can we remove the dependency to Microsoft.Data.SqlClient in Arcus.Observability.Telemetry.Sql
The Arcus Observability.Telemetry.Sql project now has a dependency on the Microsoft.Data.SqlClient package.
We use this package now just so that we can use the `SqlConnectionStringBuilder` class.  One of the `LogSqlDependency` methods takes a parameter that accepts a connection-string to a Sql database.  We use the `SqlConnectionStringBuilder` to parse this connection-string and retrieve the servername and the database-name from the connection-string.

I think it is kind of bloated and unnecessary to have a dependency on this package just for this.   Since a connection-string is nothing more then a string with multiple key/value pairs, I wonder if we cannot just write a simple 'parser' ourselves that retrieves the parts that we need.

I'm thinking of something really simple like this:

```csharp
var connectionString = "Server=myServerAddress;Database=myDataBase;User Id=myUsername;Password=myPassword;";

var parts = connectionString.Split(';')
			.Select(part => part.Split('='))
			.Where(part => part.Length == 2)
			.Select(kv => new KeyValuePair<string,string>(kv[0],kv[1])).ToArray();
```
There are of course some edge cases, as a connection string can have a `Server` component, but it is also possible to have a `Data Source` component instead.

See also the comments in this PR: https://github.com/arcus-azure/arcus.observability/pull/485

stijnmoreels:
Use our own version of parsing SQL properties instead of relying on a SQL package.

Closes #488

### DIFF
--- a/src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerSqlDependencyExtensions.cs
+++ b/src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerSqlDependencyExtensions.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using Arcus.Observability.Telemetry.Core;
 using Arcus.Observability.Telemetry.Core.Logging;
+using Arcus.Observability.Telemetry.Core.Sql;
 using GuardNet;
 
 // ReSharper disable once CheckNamespace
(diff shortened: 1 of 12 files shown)


stijnmoreels:
This also means that we have no use anymore for a dedicated package and we should slowly moving them back to the `.Core` package.

src/Arcus.Observability.Telemetry.Sql/SqlConnectionStringParser.cs
```diff
@@ -0,0 +1,72 @@
...
+                    .Reverse()
+                    .SkipWhile(char.IsWhiteSpace)
+                    .Reverse());
+        }
+
+        private static string[] GetSqlPropertyAliases(SqlProperties propertyName)
+        {
+            switch (propertyName)
+            {
+                case SqlProperties.DataSource: return new[] { "Data Source", "Server", "Addr", "Address", "Network Address" };
```
fgheysels, line 55: I was thinking of making a comment regarding the various options for retrieving the data source/server, but I hadn't seen this yet. :)
Good job.

src/Arcus.Observability.Telemetry.Sql/SqlConnectionStringParserResult.cs
```diff
@@ -0,0 +1,94 @@
+﻿using GuardNet;
+
+namespace Arcus.Observability.Telemetry.Sql
+{
+    /// <summary>
+    /// Represents the result of the <see cref="SqlConnectionStringParser"/> when parsing a SQL connection string.
+    /// </summary>
+    public class SqlConnectionStringParserResult
```
fgheysels, line 8: I agree that we should move this type to the core package, and remove the 'Sql' package.
I even think that this type can be internal ?

src/Arcus.Observability.Tests.Unit/Arcus.Observability.Tests.Unit.csproj
```diff
@@ -9,6 +9,7 @@
 
   <ItemGroup>
     <PackageReference Include="Bogus" Version="34.0.1" />
+    <PackageReference Include="Microsoft.Data.SqlClient" Version="5.0.1" />
```
fgheysels, line 12: Do we need this package here ?
(Maybe to get hold of a SqlConnection :? but since we're in unit-tests here, I wonder if that 's really necessary ?

stijnmoreels: Because we need the `SqlConnectionStringBuilder` to validate our test outcomes.

src/Arcus.Observability.Telemetry.Sql/SqlConnectionStringParserResult.cs
stijnmoreels: Could be internal, but then we wouldn't be able to test it anymore; while, as you can see, there are a lot of tests for this.

src/Arcus.Observability.Telemetry.Sql/SqlConnectionStringParser.cs
stijnmoreels: Yes, was also trying to be a bit more flexible then the original `SqlConnectionStringBuilder` where we do not fail upon invalid values of SQL properties that we do not need.

stijnmoreels:
This is actually property-based testing hiding as data-driven tests; but chose to use the latter as we should make sure that starters (which do not see property-based testing, normally) could follow.

src/Arcus.Observability.Telemetry.Sql/SqlConnectionStringParserResult.cs
fgheysels: I use `InternalsVisibleTo` for these kind of things.

src/Arcus.Observability.Tests.Unit/Arcus.Observability.Tests.Unit.csproj
fgheysels: Hm, ok.  Taught we would just do some plain old asserts since we know actually what to expect I assume.  But ok.

fgheysels (approved):

src/Arcus.Observability.Tests.Unit/Arcus.Observability.Tests.Unit.csproj
stijnmoreels: We do not exactly know as all is randomized inputs, so we would have to write a parser dedicated to the tests to find out. Therefor the `SqlConnectionStringBuilder` is a great option to make sure that the new functionality acts the same as the old.
Note that not only the key/value pairs are randomized, we also randomize whitespace, aliases, positions and character casing.

src/Arcus.Observability.Telemetry.Sql/SqlConnectionStringParserResult.cs
stijnmoreels: I know, it's just that this often feels like an anti-pattern or a sign of bad design, for me.

stijnmoreels:
We could already move the `SqlConnectionStringParser` and its result to the `.Core` package, so that our later migration is only about deprecating the extensions in the `.Sql` package and adding copies to the `.Core` package. Ok?

src/Arcus.Observability.Telemetry.Sql/SqlConnectionStringParserResult.cs
fgheysels: I think it is just a pragmatic way to make sure that you can test functionality that can be kept internal.  I have more problems with changing the visibility of a type to public, just to make it testable :)

fgheysels (approved):

src/Arcus.Observability.Telemetry.Sql/SqlConnectionStringParserResult.cs
stijnmoreels: Yeah, that's a different way of looking at things. I'm used to working with a test-driven, type-driven approach where the test is the only place first where your code is being used, and so, is already public. This was also the case here. We could make it internal but then all the tests have to be changed as well. 'Just to make it testable' is something that is often the argument of non-test-driven approaches, I guess.
I'll see if we can change this.

stijnmoreels: Made sure that the parsing is moved to the `.Core` package, made the parsing logic internal while still test the parsing via the logging extensions instead so that we do not need the `InternalsVisibleTo`.

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerSqlDependencyExtensions.cs
```diff
@@ -358,5 +359,74 @@ public static partial class ILoggerExtensions
...
+        /// <param name="connectionString">The SQL connection string.</param>
+        /// <param name="sqlCommand">The pseudo SQL command information that gets executed against the SQL dependency.</param>
+        /// <param name="operationName">The functional description of the operation that was performed on the SQL database.</param>	
+        /// <param name="isSuccessful">The indication whether or not the operation was successful.</param>
+        /// <param name="measurement">The measuring the latency to call the SQL dependency.</param>
+        /// <param name="dependencyId">The ID of the dependency to link as parent ID.</param>
+        /// <param name="context">The context that provides more insights on the dependency that was measured.</param>
+        /// <exception cref="ArgumentNullException">Thrown when the <paramref name="logger"/> is <c>null</c>.</exception>
+        /// <exception cref="ArgumentException">Thrown when the <paramref name="connectionString"/> is blank.</exception>
+        public static void LogSqlDependencyWithConnectionString(
```
fgheysels, line 376: Just wondering why the name has changed to ...WithConnectionString ?   Is it to avoid name clashes ?

fgheysels (approved):

src/Arcus.Observability.Telemetry.Sql/SqlConnectionStringParserResult.cs
fgheysels: Great thinking!

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerSqlDependencyExtensions.cs
stijnmoreels: Yeah. There was already a small clash where you would need to explicitly use the `connectionString` parameter name to void it being the `serverName` instead of other overloads; and if we would add these, we would have even more clashes. So, decided it would be better anyway to have another method name for this.

PR #515, opened 2023-01-16
feat: LogHttpDependency() should allow status codes within 100-599 range

mic-hnk:
LogHttpDependency() - allow status codes within the range of 100 to 599 - #516

### DIFF
--- a/src/Arcus.Observability.Telemetry.AspNetCore/Extensions/ILoggerHttpDependencyExtensions.cs
+++ b/src/Arcus.Observability.Telemetry.AspNetCore/Extensions/ILoggerHttpDependencyExtensions.cs
@@ -41,8 +41,6 @@ namespace Microsoft.Extensions.Logging
             Guard.NotNull(logger, nameof(logger), "Requires a logger instance to track telemetry");
             Guard.NotNull(request, nameof(request), "Requires a HTTP request message to track a HTTP dependency");
             Guard.NotNull(measurement, nameof(measurement), "Requires a dependency measurement instance to track the latency of the HTTP communication when tracking a HTTP dependency");
-            Guard.NotLessThan((int)statusCode, 100, nameof(statusCode), "Requires a valid HTTP response status code that's within the range of 100 to 599, inclusive");
-            Guard.NotGreaterThan((int)statusCode, 599, nameof(statusCode), "Requires a valid HTTP response status code that's within the range of 100 to 599, inclusive");
 
             LogHttpDependency(logger, request, statusCode, measurement.StartTime, measurement.Elapsed, context);
         }
@@ -71,8 +69,6 @@ namespace Microsoft.Extensions.Logging
             Guard.NotNull(logger, nameof(logger), "Requires a logger instance to track telemetry");
             Guard.NotNull(request, nameof(request), "Requires a HTTP request message to track a HTTP dependency");
             Guard.NotNull(measurement, nameof(measurement), "Requires a dependency measurement instance to track the latency of the HTTP communication when tracking a HTTP dependency");
-            Guard.NotLessThan((int)statusCode, 100, nameof(statusCode), "Requires a valid HTTP response status code that's within the range of 100 to 599, inclusive");
-            Guard.NotGreaterThan((int)statusCode, 599, nameof(statusCode), "Requires a valid HTTP response status code that's within the range of 100 to 599, inclusive");
 
             LogHttpDependency(logger, request, statusCode, measurement.StartTime, measurement.Elapsed, dependencyId, context);
         }
@@ -102,8 +98,6 @@ namespace Microsoft.Extensions.Logging
             Guard.NotNull(logger, nameof(logger), "Requires a logger instance to track telemetry");
             Guard.NotNull(request, nameof(request), "Requires a HTTP request message to track a HTTP dependency");
             Guard.NotLessThan(duration, TimeSpan.Zero, nameof(duration), "Requires a positive time duration of the HTTP dependency operation");
-            Guard.NotLessThan((int)statusCode, 100, nameof(statusCode), "Requires a valid HTTP response status code that's within the range of 100 to 599, inclusive");
-            Guard.NotGreaterThan((int)statusCode, 599, nameof(statusCode), "Requires a valid HTTP response status code that's within the range of 100 to 599, inclusive");
 
             context = context ?? new Dictionary<string, object>();
 
@@ -137,8 +131,39 @@ namespace Microsoft.Extensions.Logging
             Guard.NotNull(logger, nameof(logger), "Requires a logger instance to track telemetry");
             Guard.NotNull(request, nameof(request), "Requires a HTTP request message to track a HTTP dependency");
             Guard.NotLessThan(duration, TimeSpan.Zero, nameof(duration), "Requires a positive time duration of the HTTP dependency operation");
-            Guard.NotLessThan((int)statusCode, 100, nameof(statusCode), "Requires a valid HTTP response status code that's within the range of 100 to 599, inclusive");
-            Guard.NotGreaterThan((int)statusCode, 599, nameof(statusCode), "Requires a valid HTTP response status code that's within the range of 100 to 599, inclusive");
+
+            LogHttpDependency(logger, request, (int)statusCode, startTime, duration, dependencyId, context);
+        }
+
+        /// <summary>
+        /// Logs an HTTP dependency
+        /// </summary>
+        /// <param name="logger">The logger to track the telemetry.</param>
+        /// <param name="request">Request that started the HTTP communication</param>
+        /// <param name="statusCode">Status code that was returned by the service for this HTTP communication</param>
+        /// <param name="startTime">Point in time when the interaction with the HTTP dependency was started</param>
+        /// <param name="duration">Duration of the operation</param>
+        /// <param name="dependencyId">The ID of the dependency to link as parent ID.</param>
+        /// <param name="context">Context that provides more insights on the dependency that was measured</param>
+        /// <exception cref="ArgumentNullException">Thrown when the <paramref name="logger"/> or <paramref name="request"/> is <c>null</c>.</exception>
+        /// <exception cref="ArgumentOutOfRangeException">Thrown when the <paramref name="duration"/> is a negative time range.</exception>
+        /// <exception cref="ArgumentException">
+        ///     Thrown when the <paramref name="request"/> doesn't have a request URI or HTTP method, the <paramref name="statusCode"/> is outside the bounds of the enumeration.
+        /// </exception>
+        public static void LogHttpDependency(
+            this ILogger logger,
+            HttpRequest request,
+            int statusCode,
+            DateTimeOffset startTime,
+            TimeSpan duration,
+            string dependencyId,
+            Dictionary<string, object> context = null)
+        {
+            Guard.NotNull(logger, nameof(logger), "Requires a logger instance to track telemetry");
+            Guard.NotNull(request, nameof(request), "Requires a HTTP request message to track a HTTP dependency");
+            Guard.NotLessThan(duration, TimeSpan.Zero, nameof(duration), "Requires a positive time duration of the HTTP dependency operation");
+            Guard.NotLessThan(statusCode, 100, nameof(statusCode), "Requires a valid HTTP response status code that's within the range of 100 to 599, inclusive");
+            Guard.NotGreaterThan(statusCode, 599, nameof(statusCode), "Requires a valid HTTP response status code that's within the range of 100 to 599, inclusive");
 
             context = context ?? new Dictionary<string, object>();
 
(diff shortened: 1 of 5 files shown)


src/Arcus.Observability.Telemetry.AspNetCore/Extensions/ILoggerHttpDependencyExtensions.cs
```diff
@@ -41,8 +41,8 @@ public static class ILoggerHttpDependencyExtensions
             Guard.NotNull(logger, nameof(logger), "Requires a logger instance to track telemetry");
             Guard.NotNull(request, nameof(request), "Requires a HTTP request message to track a HTTP dependency");
             Guard.NotNull(measurement, nameof(measurement), "Requires a dependency measurement instance to track the latency of the HTTP communication when tracking a HTTP dependency");
-            Guard.For(() => !Enum.IsDefined(typeof(HttpStatusCode), statusCode),
-                new ArgumentException("Requires a response HTTP status code that's within the bound of the enumeration to track a HTTP dependency"));
+            Guard.NotLessThan((int)statusCode, 100, nameof(statusCode), "Requires a valid HTTP response status code that's within the range of 100 to 599, inclusive");
```
fgheysels, line 44: Thank you for your contribution!

I was thinking that we could be facing an InvalidCastException here, but it seems like that is not the case.  
(HttpStatusCode)599 for instance just works.  (This is also something that is covered by the tests).

Although, I wonder if we should have an overload that accepts an `int statusCode` to make this more explicit.  To be honest, it would not come up in my mind to just cast an arbitrary integer to a HttpStatusCode enum value, as I would have expected that this would lead to an invalidcast exception.

fgheysels (approved):

src/Arcus.Observability.Telemetry.AspNetCore/Extensions/ILoggerHttpDependencyExtensions.cs
stijnmoreels: Yes, I agree here. We have also `LogRequest` overloads that take in both `HttpStatusCode` and `int`. So it would be good to have this here, too.

stijnmoreels (changes requested):
Thx for your contribution, @mic-hnk !
Some things we have to change:
- [ ] Provide new `LogHttpDependency` overloads with `int` as status code.
- [ ] Add new tests that uses this new `int` variant.
- [ ] Updates the feature documentation to show this overload.

We can help you of course along the way!

💡 A tip for GitHub contributions: it is better to work on separate feature branches instead of on your main, otherwise conflicts and other sync problems will occur.

fgheysels:
> Thx for your contribution, @mic-hnk ! Some things we have to change:
> 
> * [ ]  Revert current changes so that the current `HttpStatusCode` overloads still validate within the enumeration.
> * [ ]  Provide new `LogHttpDependency` overloads with `int` as status code.
> * [ ]  Add new tests that uses this new `int` variant.
> * [ ]  Updates the feature documentation to show this overload.
> 
> We can help you of course along the way!
> 
> 💡 A tip for GitHub contributions: it is better to work on separate feature branches instead of on your main, otherwise conflicts and other sync problems will occur.

I don't necessarily agree with the first point @stijnmoreels , see also my comment in the issue :)

stijnmoreels:
> I don't necessarily agree with the first point @stijnmoreels , see also my comment in the issue :)

Ok, that's fine by me. Removed it. 👍

mic-hnk:
I'm on it

mic-hnk:
> 🎉 Thx for your contribution, @mic-hnk ! Greatly appreciated. Some things we have to change:
> 
> * [ ]  Provide new `LogHttpDependency` overloads with `int` as status code.
> * [ ]  Add new tests that uses this new `int` variant.
> * [ ]  Updates the feature documentation to show this overload.
> 
> We can help you of course along the way!
> 
> 💡 A tip for GitHub contributions: it is better to work on separate feature branches instead of on your main, otherwise conflicts and other sync problems will occur.

I pushed some changes.
Also, I guess this is the documentation you mean - https://observability.arcus-azure.net/features/writing-different-telemetry-types/#measuring-http-dependencies
Do you have any suggestion how I should indicate difference between previously existing methods and the `int` overload?

fgheysels:
> > 🎉 Thx for your contribution, @mic-hnk ! Greatly appreciated. Some things we have to change:
> > 
> > * [ ]   Provide new `LogHttpDependency` overloads with `int` as status code.
> > * [ ]   Add new tests that uses this new `int` variant.
> > * [ ]   Updates the feature documentation to show this overload.
> > 
> > We can help you of course along the way!
> > 💡 A tip for GitHub contributions: it is better to work on separate feature branches instead of on your main, otherwise conflicts and other sync problems will occur.
> 
> I pushed some changes. Also, I guess this is the documentation you mean - https://observability.arcus-azure.net/features/writing-different-telemetry-types/#measuring-http-dependencies Do you have any suggestion how I should indicate difference between previously existing methods and the `int` overload?

To be honest, I think the docs as is is sufficient.   In my opinion, a developer will be able to detect this overload by using intellisense.

stijnmoreels:
> To be honest, I think the docs are sufficient as is. In my opinion, a developer will be able to detect this overload by using intellisense.

Just looked, you might be right 🤔 . We can leave it like this. 👍

src/Arcus.Observability.Telemetry.AspNetCore/Extensions/ILoggerHttpDependencyExtensions.cs
```diff
@@ -137,16 +131,47 @@ public static class ILoggerHttpDependencyExtensions
...
+        /// <param name="statusCode">Status code that was returned by the service for this HTTP communication</param>
+        /// <param name="startTime">Point in time when the interaction with the HTTP dependency was started</param>
+        /// <param name="duration">Duration of the operation</param>
+        /// <param name="dependencyId">The ID of the dependency to link as parent ID.</param>
+        /// <param name="context">Context that provides more insights on the dependency that was measured</param>
+        /// <exception cref="ArgumentNullException">Thrown when the <paramref name="logger"/> or <paramref name="request"/> is <c>null</c>.</exception>
+        /// <exception cref="ArgumentOutOfRangeException">Thrown when the <paramref name="duration"/> is a negative time range.</exception>
+        /// <exception cref="ArgumentException">
+        ///     Thrown when the <paramref name="request"/> doesn't have a request URI or HTTP method, the <paramref name="statusCode"/> is outside the bounds of the enumeration.
+        /// </exception>
```
stijnmoreels, line 152:
```suggestion
        /// <exception cref="ArgumentException">
        ///     Thrown when the <paramref name="request"/> doesn't have a request URI or HTTP method, the <paramref name="statusCode"/> is outside the 100-599 range inclusively.
        /// </exception>
```

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerHttpDependencyExtensions.cs
```diff
@@ -164,18 +160,49 @@ public static partial class ILoggerExtensions
...
+        /// <param name="statusCode">The status code that was returned by the service for this HTTP communication.</param>
+        /// <param name="startTime">The point in time when the interaction with the HTTP dependency was started.</param>
+        /// <param name="duration">The duration of the operation.</param>
+        /// <param name="dependencyId">The ID of the dependency to link as parent ID.</param>
+        /// <param name="context">The context that provides more insights on the dependency that was measured.</param>
+        /// <exception cref="ArgumentNullException">Thrown when the <paramref name="logger"/> or <paramref name="request"/> is <c>null</c>.</exception>
+        /// <exception cref="ArgumentOutOfRangeException">Thrown when the <paramref name="duration"/> is a negative time range.</exception>
+        /// <exception cref="ArgumentException">
+        ///     Thrown when the <paramref name="request"/> doesn't have a request URI or HTTP method, the <paramref name="statusCode"/> is outside the bounds of the enumeration.
+        /// </exception>
```
stijnmoreels, line 181:
```suggestion
        /// <exception cref="ArgumentException">
        ///     Thrown when the <paramref name="request"/> doesn't have a request URI or HTTP method, the <paramref name="statusCode"/> is outside the 100-599 range inclusively.
        /// </exception>
```

src/Arcus.Observability.Telemetry.AspNetCore/Extensions/ILoggerHttpDependencyExtensions.cs
```diff
@@ -137,16 +131,47 @@ public static class ILoggerHttpDependencyExtensions
...
+        public static void LogHttpDependency(
```
stijnmoreels, line 153: This is very good! 👍 
We also need a variant that takes in the `DurationMeasurement` instead of the `DateTimeOffset`/`TimeSpan` pair.

src/Arcus.Observability.Telemetry.Core/Extensions/ILoggerHttpDependencyExtensions.cs
```diff
@@ -164,18 +160,49 @@ public static partial class ILoggerExtensions
...
+        public static void LogHttpDependency(
```
stijnmoreels, line 182: This is very good! 👍 
We also need a variant that takes in the `DurationMeasurement` instead of the `DateTimeOffset`/`TimeSpan` pair.

src/Arcus.Observability.Tests.Unit/Telemetry/Logging/HttpDependencyLoggingTests.cs
```diff
@@ -417,7 +417,7 @@ public void LogHttpDependencyWithRequestWithDurationMeasurementWithDependencyId_
             // Arrange
             var logger = new TestLogger();
             HttpRequest request = CreateStubRequest(HttpMethod.Get, "host", "/path", "http");
-            var statusCode = BogusGenerator.PickRandom<HttpStatusCode>();
+            var statusCode = (HttpStatusCode)BogusGenerator.Random.Int(100, 599);
```
stijnmoreels, line 420: Good thing that these tests are adapted so that we can be sure that the guard for `HttpStatusCode` is removed. 👍

src/Arcus.Observability.Tests.Unit/Telemetry/Logging/HttpDependencyLoggingTests.cs
```diff
@@ -426,6 +426,38 @@ public void LogHttpDependencyWithRequestWithDurationMeasurementWithDependencyId_
             Assert.ThrowsAny<ArgumentException>(() => logger.LogHttpDependency(request, statusCode, measurement: null, dependencyId));
         }
 
+        [Theory]
+        [InlineData(99)]
+        [InlineData(600)]
```
stijnmoreels, line 431: To be even more secure (and without the lovely property-based testing), we should be more general in generating inputs here.

This will give us a bit more security:
```csharp
int statusCode = BogusGenerator.PickRandom(
    BogusGenerator.Random.Int(max: 99), 
    BogusGenerator.Random.Int(min: 600));
```

stijnmoreels (changes requested):
Thank you so much for taking this up! 🥇
Some small comments but were getting there.

src/Arcus.Observability.Telemetry.AspNetCore/Extensions/ILoggerHttpDependencyExtensions.cs
mic-hnk: done

src/Arcus.Observability.Tests.Unit/Telemetry/Logging/HttpDependencyLoggingTests.cs
mic-hnk: done

src/Arcus.Observability.Telemetry.AspNetCore/Extensions/ILoggerHttpDependencyExtensions.cs
stijnmoreels: Yes, so that we both have a `DateTimeOffset`/`TimeSpan` signature and a `DurationMeasurement` signature for both `HttpStatusCode` and `int`. 👍

stijnmoreels (approved):
Think that this new variant you added also needs some unit tests. But otherwise this is good to go! 💯 
Great job re-using the common code practices and helping us bringing higher value! Greatly appreciated. 💗

mic-hnk:
> Think that this new variant you added also needs some unit tests. But otherwise this is good to go! 💯 Great job re-using the common code practices and helping us bringing higher value! Greatly appreciated. 💗

I added some more tests :)

fgheysels:
Thanks for this contribution.
This is a small but convenient change which will improve Arcus' usability!

mic-hnk:
Thank you :)
This project provides a lot of value and I'm glad it's there. Keep up the good work!

PR #537, opened 2023-08-17
chore: add net8.0 preview support

stijnmoreels:
Add .NET 8 (preview) support to the project.

Relates to https://github.com/arcus-azure/arcus/issues/311


stijnmoreels:
Waiting for https://github.com/arcus-azure/azure-devops-templates/pull/57 to include preview .NET SDK's in the build.

src/Arcus.Observability.Tests.Runtimes.AzureFunction/Dockerfile
```diff
@@ -4,7 +4,7 @@ FROM mcr.microsoft.com/azure-functions/dotnet:4.0 AS base
 WORKDIR /home/site/wwwroot
 EXPOSE 80
 
-FROM mcr.microsoft.com/dotnet/sdk:6.0.100-bullseye-slim AS build
+FROM mcr.microsoft.com/dotnet/sdk:8.0-preview-bullseye-slim AS build
```
fgheysels, line 7: If I'm not mistaken, this container is used to run the tests inside a container.
is there a possibility to run these tests for multiple .NET targets ?

Like, maybe parametrize/tokenize the base-images in the dockerfile, and build the container multiple times in the pipeline, using different base images ?

stijnmoreels: Yes, that could be possible, maybe with a YAML loop for base images and multiple test suite runs for the Docker-specific integration tests.
Though, not sure if this should be a blocking factor here as this is probably something that we will be using across multiple repo's (plus, my Docker skills are very very poor 😄).

stijnmoreels:
Azure Functions v4 is not yet available for .NET 8?
https://learn.microsoft.com/en-us/azure/azure-functions/functions-versions?tabs=v4&pivots=programming-language-csharp#languages

fgheysels (approved):

src/Arcus.Observability.Tests.Runtimes.AzureFunction/Dockerfile
```diff
@@ -13,7 +13,7 @@ WORKDIR "/src/Arcus.Observability.Tests.Runtimes.AzureFunction"
 RUN dotnet build "Arcus.Observability.Tests.Runtimes.AzureFunction.csproj" -c Release -o /app/build
 
 FROM build AS publish
-RUN dotnet publish "Arcus.Observability.Tests.Runtimes.AzureFunction.csproj" -c Release -o /app/publish
+RUN dotnet publish "Arcus.Observability.Tests.Runtimes.AzureFunction.csproj" -c Release -o /app/publish /p:UseAppHost=false
```
fgheysels, line 16: I'm wondering why we need this.

I've been reading a bit the docs on this option, but I'm afraid I don't really understand :)

I understand that by using this option, we're falling back on the 'framework dependent deployment' whereas since .NET 3, 'framework dependent executable' is the default mode.
With FDD, a dll is generated and that dll is started, whereas with FDE, an exe is created.   Why do we need to fallback to something that is not the default mode anymore ?

fgheysels (commented):
So, you succeeded in getting the dockerstuff working \o/

fgheysels:
There are some hints / advise that the code-analysis is reporting regarding the usage of obsolete methods.  MIght be good to tackle that as well ?

src/Arcus.Observability.Tests.Runtimes.AzureFunction/Dockerfile
stijnmoreels: Honestly, this is close to Chinese to me most of the time, no matter how much a spend time in the Docker-Wonderworld. To fix the Docker issues, I started small again (minimal), with the default contents that are in the Microsoft project templates. This option was there, so I left it, but if you say this is not what we want; we can off course remove it.

fgheysels: I wonder if it breaks if you remove it :)

stijnmoreels:
> There are some hints / advise that the code-analysis is reporting regarding the usage of obsolete methods. MIght be good to tackle that as well ?

Think we currently report the code coverage and any security risks. Obsolete functionality should indeed be part of that security risk assessment. But: we have enabled 'warnings as errors' so those should also break the build.

src/Arcus.Observability.Tests.Runtimes.AzureFunction/Dockerfile
stijnmoreels: Let's try.

stijnmoreels: It works 👍🎉

Issue #545, opened 2023-10-18
Log Custom Event is unnecessarily enriching source Context


papabytes:
**Describe the bug**
When we call LogCustomEvent the code in [EventLogEntry](src/Arcus.Observability.Telemetry.Core/Logging/EventLogEntry.cs) line 25 adds a new Key Value Pair to a Context dictionary that should not be altered on the original scope of the code.
In a scenario that we have, we're sending a message to a ServiceBus Queue  and passing the current contex into this method. After the LogCustomEvent call our context is enriched with an unsolicited property: TelemetryType.

We use our source dictionary to enrich the properties of the ServiceBusMessage. This TelemetryType is not serializable resulting in an exception that prevents the message to be sent to the Service Bus

**To Reproduce**
Create a ServiceBus Message with custom Properties sourcing from a local Dictionary<string,object>.
Call LogCustomEvent with those properties.
Set the ServiceBusMessage Custom properties to your local Dictionary.
Check that the Message fails to be sent as ContextProperties.General.TelemetryType is not Serializable

**Expected behavior**
The local Context should not be manipulated by the Arcus Observability LogCustomEvent Logic.
If you want to take the passed context as base, please initalize a new Dictionary from the one provided i.e. on line 24
`
Context = context ? new Dictionary(context) :  new Dictionary<string, object>();
`

stijnmoreels:
I can see your point, and it is indeed something that would occur if the telemetry context is reused across multiple places, instead of creating fresh ones at each point.
This is, though, something that has to be fixed across all log extensions.
@fgheysels , you agree?

@papabytes , would you care to take this up? The description in the 'Reproduce' could be a guideline test to remove this everywhere.
Thx for mentioning!

stijnmoreels:
If you do not have the time, we're happy to pick this up and let it be part of our upcoming v3 version! 👍

fgheysels:
I agree, we should not make modifications to any object that is passed in.
In this case, we're running into a side effect since the Dictionary is a reference type and is passed in by reference.   Since we're adding a new entry to our Context, this is actually also being added in the passed in dictionary, since it has been added by ref.

We can solve it as it has been proposed by @papabytes , but maybe we can also solve it by specifying the `in` parameter modifier ? (https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/in-parameter-modifier)
Unless that would result in a compiler error.

stijnmoreels:
[...]

PR #546, opened 2023-10-24
fix: do not alter original telemetry context pt.1

stijnmoreels:
Fixes the first set of telemetry extensions w/ a telemetry context that does not alter the original passed-in context. Split in two parts as there are a lot of extensions/tests to be changed.

Relates to #545


fgheysels:
I was first wondering if we would have a test that verifies if the original context is not altered, and after reviewing some files, I indeed realized  :  "why should I even wonder" .  Thanks for adding the additional test! :)

fgheysels (approved):

stijnmoreels:
> I was first wondering if we would have a test that verifies if the original context is not altered, and after reviewing some files, I indeed realized : "why should I even wonder" . Thanks for adding the additional test! :)

Also simplified the telemetry context determination to a single location, so that it becomes easier to see where the context is 'copied from' or 'created anew'.
The test are mostly verifying these indeed, but I could have missed an overload in the process; hence why I simplified it, so that we know for sure that we're testing the same thing.

PR #548, opened 2023-10-27
feat: add managed identity support to app insights serilog sink

### LINKED ISSUE #541: Azure AD Authentication In the ApplicationInsights Sink
Azure provides now a way to authenticate towards AppIns via a service principal. I was always suprised why they didn't provide something like this sooner, but hey :).

Now it is supported I want to leverage from this via Arcus.
When setting up your sink, there should be a way to set this authentication.
More information on the tool itself: https://learn.microsoft.com/en-us/azure/azure-monitor/app/azure-ad-authentication?tabs=net.

There are two ways we can handle these things...

We now have in the sink startup `WriteTo.AzureApplicationInsightsWithConnectionString`, where we could add an overload parameter with tokenCredential (so you can choose managed identity or service principal).

Or we add an extension method on top of `WriteTo.AzureApplicationInsightsWithConnectionString`, with name similar to `'UsingAuthenication('TokenCredential credential)`.

**Describe alternatives you've considered**
There are no real alternatives, I am a bit stuck as I cannot use the Serilog sink as it doesn't do the convertions.

**Additional context**
/

stijnmoreels:
Adds managed identity support to the Serilog Application Insights sink that lets you provide a client ID together with the ingestion connection string towards Azure Application Insights.

Closes #541

### DIFF
--- a/docs/preview/03-Features/sinks/azure-application-insights.md
+++ b/docs/preview/03-Features/sinks/azure-application-insights.md
@@ -11,42 +11,50 @@ The Azure Application Insights sink is an extension of the [official Application
 
 You can easily configure the sink by providing the Azure Application Insights connection string or instrumentation key, but the connection string is preferred.
 
-```csharp
-using Serilog;
-using Serilog.Configuration;
-
-ILogger logger = new LoggerConfiguration()
-    .MinimumLevel.Debug()
-    .WriteTo.AzureApplicationInsightsWithConnectionString("<connection-string>")
-    .CreateLogger();
-
-ILogger logger = new LoggerConfiguration()
-    .MinimumLevel.Debug()
-    .WriteTo.AzureApplicationInsightsWithInstrumentationKey("<key>")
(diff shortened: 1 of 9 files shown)


joachimgoris:
Is this feature still wanted after being stale for 2 years? Otherwise I could pick this up 😄

stijnmoreels:
> Is this feature still wanted after being stale for 2 years? Otherwise I could pick this up 😄

First some slimmed-down exercise needs to happen on the Arcus repo's before we can start with new functionality, I would say.

joachimgoris:
Sure. How can I help? I presume you'd first want to do an analysis?

stijnmoreels:
> Sure. How can I help? I presume you'd first want to do an analysis?

Yes, working on that here: https://github.com/arcus-azure/arcus.observability/discussions/578
But first other repo's are prioritized, I believe.

joachimgoris:
I can also help out in any other repo.

PR #572, opened 2024-04-17
full package upgrades

laurensvc:


fgheysels (approved):

stijnmoreels:
Closed in favor of #597 where we use local search and PNPM instead.

PR #574, opened 2024-05-30
chore: deploy/use new test resources in dev tenant

stijnmoreels:
* Adds a new pipeline to deploy test resources to the new Dev tenant,
* Alters the current CI/Release pipelines to pick any secrets from a vault, deployed with the test resources.

### DIFF
--- a/build/deploy-test-resources.yml
+++ b/build/deploy-test-resources.yml
@@ -41,14 +41,15 @@ stages:
 
                 $parameters = @{
                   "location" = "westeurope"
-                  "appInsightsName" = "${{ parameters.appInsightsName }}" } | ConvertTo-Json
+                  "appInsightsName" = ${{ parameters.appInsightsName }} } | ConvertTo-Json
                 $appInsightsOutput = az deployment sub create `
                   --location westeurope `
                   --template-file ./build/templates/deploy-test-resources.bicep `
                   --parameters $parameters `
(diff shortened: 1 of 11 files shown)


build/templates/deploy-test-resources.bicep
```diff
@@ -0,0 +1,71 @@
...
+    secrets: [
+      {
+        name: 'ApplicationInsights-InstrumentationKey'
+        value: component.outputs.instrumentationKey
+      }
+    ]
+  }
+}
+
+output ApplicationInsights_WorkspaceId string = workspace.outputs.resourceId
```
stijnmoreels, line 70: Will be used when we migrate to the new app insights query nuget package.

build/deploy-test-resources.yml
```diff
@@ -0,0 +1,83 @@
...
+    displayName: 'Variable group name'
+    default: Arcus Observability - Integration Testing
+
+variables:
+  - name: keyVaultName
+    value: 'arcus-observability-kv'
+  - template: ./variables/build.yml
+
+stages:
+  - stage: Deploy
```
stijnmoreels, line 27: Best that we also add the self-contained integration tests as a second depend stage, so that we always have a verification that the deployment went successful. Since they are not expensive, we can use them as some kind of smoke tests.

PR #576, opened 2024-07-30
chore: use new monitor package w/ newly deployed resources

stijnmoreels:
Switches to the new Azure package that let's you query telemetry, and removing the older, deprecated one. This, so that we do not have to rely on the (also) deprecated API key authentication when querying telemetry.
PR also includes the Bicep where the updated tests interact with.


stijnmoreels:
@fgheysels this both fixes the usage of the deprecated monitor package as well as deploy to the new Dev tenant.

fgheysels (approved):

PR #600, opened 2025-06-10
chore(deprecate): logging extensions w/ logging scope

stijnmoreels:
Logging scopes are supported in OpenTelemetry as well, we have some methods that starts the log scope, but since it's recommended to use another type (`IReadOnlyList<KeyValue<string, object?>>`) AND the implemented code is too little to provide value while it has unnecessary support costs (see also discussion #587), this PR deprecates these extensions.


stijnmoreels:
> I even wonder if we need the check with .IsEnabled on the loglevel. If we just call Log with the desired LogLevel, and the configured loglevel is higher than the one passed in, I believe that trace won't be logged ?

That's what I was saying: that checking this is even a bad .NET practice regarding performance.

> I see what you mean, but I think it's a handy shortcut to provide additional context to your log-trace without you having to call BeginScope yourself and clutter your codebase.

Yes, true, it is handy, but that puts a burden on us again, regarding maintainence, while most people will probably use the Microsoft documentation when using logging scopes.

fgheysels (approved):
Ok, if we need to add additional context, we can start a loggingscope ourselves.
It's also possible (built-in) to add additional context via placeholders (don't know the exact name by heart)

```
_logger.LogInformation("Some info for entity {EntityId} and other {Stuff}", entityId, stuff);
```

PR #605, opened 2025-08-21
chore(deprecate): application name/version enrichment for being too project-specific

stijnmoreels:
As mentioned in Arcus sessions and documented at #587 , the application name/versioning enrichment is too project-specific and a bit outside the scope of what Arcus should provide.

This PR deprecates this application name/versioning functionality.

### DIFF
--- a/src/Arcus.Observability.Correlation/Arcus.Observability.Correlation.csproj
+++ b/src/Arcus.Observability.Correlation/Arcus.Observability.Correlation.csproj
@@ -25,7 +25,6 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Guard.Net" Version="3.0.0" />
     <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="6.0.0" />
     <PackageReference Include="Microsoft.Extensions.Options" Version="6.0.0" />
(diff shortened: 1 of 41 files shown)


src/Arcus.Observability.Telemetry.Core/AssemblyAppVersion.cs
```diff
@@ -1,13 +1,15 @@
...
 
 namespace Arcus.Observability.Telemetry.Core
 {
     /// <summary>
     /// Represents an <see cref="IAppVersion"/> implementation that uses the assembly version as application version.
     /// </summary>
     /// <seealso cref="IAppVersion"/>
+#pragma warning disable S1133
+    [Obsolete("Will be removed in v4.0 as application versioning enrichment is too project-specific")]
+#pragma warning restore S1133
```
fgheysels, line 12: I don't think we should do that.  IMHO, this is really valuable.

stijnmoreels: Maybe in the older Serilog setup, but with OpenTelemetry these 'resource information' is set differently and on different places, as well. Think their way of working is rather clean and interfaces like these would maybe overcomplicate it.

